PR #7, opened 2021-04-22
build: add vue and react to package.exports

euaaaio:
Looks like we need it at least for Vite.


ai:
You are right. We need it also for Node.js 16.

PR #12, opened 2021-05-12
preact support

Merciful12:


package.json
```diff
@@ -73,6 +74,7 @@
     "jest-localstorage-mock": "^2.4.12",
     "lint-staged": "^11.0.0",
     "nanodelay": "^2.0.0",
+    "preact": "^10.5.13",
```
ai, line 77: Do we need to add optional peer dependency as we did for `react`?

ai:
Also, what is your Twitter&Telegram accounts to mention you in releae tweet?

package.json
Merciful12: I think yes, but not sure which version it should be

Merciful12:
> Also, what is your Twitter&Telegram accounts to mention you in release tweet?

Twitter: `@12Future12`
Telegram: `@Merciful`

ai:
Is it OK if I will release it a few days later? You do not pan to use it soon, don't you?

ai:
To have full Preact support we need to port this fles too https://github.com/logux/client/tree/main/react

PR #15, opened 2021-05-29
Improve docs 

euaaaio:


ai:
Thanks. I was worried about my Vue examples.

euaaaio:
> Thanks. I was worried about my Vue examples.

It ok. Your example can be easily fixed by Vue errors in console.

PR #20, opened 2021-06-01
feat: make set method unavailable on router instance

Guria:
Calling `set` method directly could cause state not in sync with actual location.
This includes both fix in types and in runtime.
Introduced new generic type `StoreLike` with minimal subset of properties most utility functions expect to work with.

### DIFF
--- a/create-router/index.js
+++ b/create-router/index.js
@@ -70,12 +70,12 @@ export function createRouter(routes) {
 
   let popstate = () => {
     let page = parse(location.pathname)
-    if (page !== false) router.set(page)
+    if (page !== false) setState(page)
   }
 
   let router = createStore(() => {
     let page = parse(location.pathname)
-    if (page !== false) router.set(page)
+    if (page !== false) setState(page)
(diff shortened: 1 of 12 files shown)


create-router/index.js
```diff
@@ -85,13 +85,15 @@ export function createRouter(routes) {
     }
   })
 
+  let setState = router.set
+  delete router.set
```
Guria, line 89: We can wrap delete statement with non production env check. It will save few bytes.
Current version add 11 B, with check it would be only 4. Not sure if it makes sense or not.
@ai

ai:
ESLint issue need to be fixed.

Guria:
@ai fixed

ai:
I find a better way d7735f7

1. More clear type names: `ReadableStore`/`WritableStore`
2. I hide `router.set` only in `development` to save size
3. `?.` removal was wrong. It was an error in `ReadonlyIfObject` type.

But thanks for raising the problem of read-only stores.

Guria:
Perfect

PR #22, opened 2021-06-02
refactor(create-router): decouple DOM specific methods into factory

Guria:
All DOM specific API isolated and could be moved into own module or package alongside with DOM specific tests

Bringing this for discussion purposes first of all.
I'd like to ask a permission to extract `bindRouterEvents` as separate repo/module.
I'm thinking about making a collection of small modules helpers that allow to build own routing solutions.

Unfortunately it brings extra 44 B, but it reduces responsibility of router module.

### DIFF
--- a/create-router/index.js
+++ b/create-router/index.js
@@ -1,5 +1,64 @@
 import { createStore } from '../create-store/index.js'
 
+function bindRouterEvents(parse, setValue, element = document.body) {
+  let click = event => {
+    let link = event.target.closest('a')
+    if (
+      !event.defaultPrevented &&
+      link &&
+      event.button === 0 &&
+      link.target !== '_blank' &&
+      link.dataset.noRouter == null &&
+      link.rel !== 'external' &&
+      !link.download &&
+      !event.metaKey &&
+      !event.ctrlKey &&
+      !event.shiftKey &&
+      !event.altKey
+    ) {
+      let url = new URL(link.href)
+      if (url.origin === location.origin) {
+        event.preventDefault()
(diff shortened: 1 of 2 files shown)


Guria:
@ai let me know your thoughts on it

Guria:
So 2 questions here:
* does it make sense to have this function decoupled still being part of this project
* would you mind if I extract this part as separate module anyway?

ai:
> does it make sense to have this function decoupled still being part of this project

Does it have any rational reason to excuse size increasing

> would you mind if I extract this part as separate module anyway?

Feel free to copy it to your project

Guria:
> Does it have any rational reasons to excuse size increasing?

Separations of responsibilities and concerns only. We will have all DOM API related parts along with tests decoupled and allow router module to concentrate on parsing/stringifying and integration with store.
Additionally If someone decide to have own version of router (eg, to add custom query/hash parsers like [urlon](https://github.com/cerebral/urlon) or [qs](https://github.com/ljharb/qs)) it would be easier to start with decoupled version as base.

I will let you to decide whether it rational enoug excuse or not.

ai:
> Separations of responsibilities and concerns only

It is reasons for us, but not for end-user. We should increase the size only if it improves end-user (developers) life.

Guria:
@ai I've started extracting for my personal needs and eventually came up with much better results both in terms of API and size-limit.
Now it is only 5 bytes extra if we keep current API for router.open. But we can even save few bytes if we will make it kinda "private".

PR #31, opened 2021-06-13
Reduce size and fix docs of Vue part

euaaaio:
It can wait for the next release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -188,7 +188,7 @@
         "./vue/index.js": "{ useStore }",
         "./index.js": "{ createStore, defineMap, createDerived, getValue }"
       },
-      "limit": "560 B"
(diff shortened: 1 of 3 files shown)


ai:
Do we also need to fix https://github.com/logux/client/blob/main/vue/index.js#L106 ?

euaaaio:
> Do we also need to fix https://github.com/logux/client/blob/main/vue/index.js#L106 ?

Yes. Now we can. I’ll make a PR.

As far as I know, there are no best practices for using `readonly`. But maybe we can use `readonly` for non-production?

ai:
> But maybe we can use readonly for non-production?

Yeap. I like the idea to cut checks and protections from production code.

PR #36, opened 2021-07-01
Improve spelling in README

vasilii-kovalev:
* Change "it's" to "it is" to make the text consistent
* Use definite and Indefinite articles in some places (where I think they should be IMO)
* Add "-s" ending to "listen" when used with a single form of a noun
* Change "place" to "source" in "single source of truth", since it is an idiom


ai:
Can you change PR to send changes to `next` branch?

README.md
```diff
@@ -170,7 +170,7 @@ You can change store value by calling the `store.set(newValue)` method.
 
 ### Map Store
 
-This store with key-value pairs.
+This store is with key-value pairs.
```
ai, line 173:
```suggestion
The store with key-value pairs.
```

`is with` can't be used in English

PR #46, opened 2021-08-08
Feature: add observe-only option to use model

droganov:


.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules/
 yarn-error.log
 
 coverage/
+.tool-versions
```
ai, line 5: You need to move it to your local git ignore
```suggestion
```

react/index.d.ts
```diff
@@ -1,5 +1,9 @@
 import { ReadableStore } from '../create-store/index.js'
 
+export interface UseStoreOptions {
+  observeOnly?: string[]
```
ai, line 4: Too long and hard to remember.

What do you think about `only`?

react/index.d.ts
```diff
@@ -23,7 +27,10 @@ import { ReadableStore } from '../create-store/index.js'
  * @param store Store instance.
  * @returns Store value.
  */
-export function useStore<Value extends any>(store: ReadableStore<Value>): Value
+export function useStore<Value extends any>(
+  store: ReadableStore<Value>,
+  options?: UseStoreOptions
+): Value
```
ai, line 33: On `only` option, we need to modify returning `Value` to avoid using keys which was not `only` options

react/index.d.ts
```diff
@@ -23,7 +27,10 @@ import { ReadableStore } from '../create-store/index.js'
  * @param store Store instance.
  * @returns Store value.
  */
-export function useStore<Value extends any>(store: ReadableStore<Value>): Value
+export function useStore<Value extends any>(
```
ai, line 30: We need to change Preact version too

react/index.d.ts
droganov: > Too long and hard to remember.
> 
> What do you think about `only`?

only is too uncertain, maybe keys?

react/index.d.ts
droganov: We have no filter inside the hook, we only mute events...
Should we filter return value as well?

react/index.d.ts
ai: :+1:

react/index.d.ts
ai: I'm afraid that somebody will forget to update `keys` on changing the React template and then will spend a day debugging React render. We do not need to change returning value, just change type to help user

PR #52, opened 2021-08-30
Allow keys of every union part to be used in `changedKey`

AleksandrSl:
Details: https://stackoverflow.com/a/61685380/6540091
This should not lead to any additional errors, because `changedKey` is usually used in comparisons.

### DIFF
--- a/create-map/index.d.ts
+++ b/create-map/index.d.ts
@@ -1,3 +1,5 @@
+type AllKeys<T> = T extends any ? keyof T : never
+
 export interface MapStore<Value extends object = any> {
   /**
    * `true` if store has any listeners.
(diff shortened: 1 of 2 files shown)


create-map/index.d.ts
```diff
@@ -103,3 +103,5 @@ export interface MapStore<Value extends object = any> {
 export function createMap<Value extends object, StoreExt extends object = {}>(
   init?: () => void | (() => void)
 ): MapStore<Value> & StoreExt
+
+type AllKeys<T> = T extends any ? keyof T : never
```
ai, line 107: Can you move it before the first usage? Just to improve readability.

ai:
Thanks. Released in 0.4.4

PR #53, opened 2021-08-31
fix: union type in createMap

Beraliv:
* Fixed `setKey` and `notify`
* Also fixed `value` in `setKey`


ai:
Thanks. The fix was released in 0.4.5.

PR #58, opened 2021-09-10
fix: diamond problem

Eddort:
This pull request resolves the diamond dependency problem in 2 steps:
1. Listen to only writable stores
2. Bind callback with arguments (**run** function)


create-derived/index.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/restrict-plus-operands */
```
ai, line 1: You do not need to disable this ESLint rule. Just use:

```js
    return `a${count}`
```

ai:
1. Can I ask you to remove Diamond Problem note from the docs? I will make a link to PR in Twitter, it will be nice to fix the docs here.
2. Can you update Size Limit config and max size at the beginning of the docs? The size increase is low, great work. Just to make PR green to promote in Twitter.

create-derived/index.js
```diff
@@ -1,20 +1,43 @@
...
+  )
+]
 
 export function createDerived(stores, cb) {
   if (!Array.isArray(stores)) stores = [stores]
+  let deps = collectWritable(stores)
+
+  let run = () => {
+    let values = stores.map(store =>
+      store.wrapper ? store.wrapper() : getValue(store)
```
ai, line 19: Who can set `store.wrapper`?

create-derived/index.test.ts
Eddort: Sure, it was a prototype

create-derived/index.js
Eddort: Resolved. I succeed to reduce the additional size from 54 to 35 B

ai: 😍

ai (approved):

ai:
Looks good. Will merge it when I will finish unpacking (relocation is hard 😅).

PR #60, opened 2021-09-13
Removed create* prefix

usmanyunusov:


ai:
We need to keep backward compatibility. We need to copy old methods to `depreacted/index.js` and `deprecated/index.d.ts`?

usmanyunusov:
@ai, Added only createMap to depreacted

PR #62, opened 2021-09-19
Nanostores lifecycle

Eddort:
PR is in progress right now. I need more time for refactoring. But it's working :)


Eddort:
I'm ready.
What I’ve done:
1. Rewritten `map`, `mapTemplate`, and `atom` and all places where they were used.
2. Written `lifecycle` API
3. Added `mount` support
4. Moved to deprecated `getValue` function.
5. Provided backward compatibility (`deprecated` file)

ai:
🎉🤯

Sorry, I am a few emotional moments this weekend. I will deeply review in on next week.

atom/depricated.d.ts
```diff
@@ -0,0 +1,101 @@
...
+ *   return () => {
+ *     window.removeEventListener('popstate', parse)
+ *   }
+ * })
+ * ```
+ *
+ * @param init Initialize store and return store destructor.
+ * @returns The store object with methods to subscribe.
+ */
+export function atom<Value, StoreExt = {}>(
```
ai, line 99: You do not need to add `atom` to deprecation, since we never released it. You can remove it.

atom/index.d.ts
```diff
@@ -1,12 +1,4 @@
-export const STORE_CLEAN_DELAY: number
-
-type ReadonlyIfObject<Value> = Value extends undefined
-  ? Value
-  : Value extends (...args: any) => any
-  ? Value
-  : Value extends object
-  ? Readonly<Value>
-  : Value
+type ReadonlyIfObject<Value> = Value extends object ? Readonly<Value> : Value
```
ai, line 1:
```ts
type B = ReadonlyIfObject<(a: number) => number>
```

`B` will be `{}` type. We need this check for `function`

ai: https://www.typescriptlang.org/play?#code/C4TwDgpgBAglC8UAUBDAXFAdgVwLYCMIAnASgQD4s9CiAoW0SKAJQhQBMB7TAGxAEkAZgHl8AKwgBjYAB4Aaih7YIlRAqXQIAD2ARM7AM5RO4qcCgB+Fmy68Q8xcsoZ1y+o2gAhBNY7c+QqIS0jKoGDgExGTwlBE05EA

atom/index.d.ts
```diff
@@ -96,6 +94,4 @@ export interface WritableStore<Value = any> extends ReadableStore<Value> {
  * @param init Initialize store and return store destructor.
  * @returns The store object with methods to subscribe.
  */
-export function atom<Value, StoreExt = {}>(
-  init?: () => void | (() => void)
-): WritableStore<Value> & StoreExt
+export function atom<Value>(state?: Value): WritableStore<Value>
```
ai, line 97: Why did we remove `StoreExt`  type parameter?

deprecated/index.js
```diff
@@ -28,5 +30,10 @@ export function defineMap(...args) {
 
 export function createMap(...args) {
   warning('Replace createMap() to map()')
-  return map(...args)
+  return mount(map(), ...args)
+}
+
+export function getValue(...args) {
+  warning('Replace getValue(store) to store.get()')
+  return getValueDeprecated(...args)
```
ai, line 38: Let’s move all deprecated code to this file.

For instance, move `getValue`.

get-value/index.js
```diff
@@ -1,12 +1,3 @@
...
-    return store.value
-  } else {
-    let result
-    let unbind = store.subscribe(value => {
-      result = value
-    })
-    unbind()
-    return result
-  }
+  return store.get()
```
ai, line 2: Better to move it to `depreacted/index.js`

lifecycle/index.d.ts
```diff
@@ -0,0 +1,35 @@
...
+    original: unknown[]
+    shared: any
+    stop(): void
+    abort(): void
+  }) => void
+)
+
+export function onCreate<Data>(
+  store: Store<Data>,
+  handler: (payload: { shared: any }) => void
```
ai, line 27: We can avoid `any` by asking developer to clarify what previous event handlers put there:

```ts
export function onCreate<Shared = never>(
  store: Store,
  handler: (payload: { shared: Shared }) => void
)
```

lifecycle/index.d.ts
```diff
@@ -0,0 +1,35 @@
+import type { WritableStore, ReadableStore } from '../atom/index.js'
+
+type Store<Data> = WritableStore<Data> | ReadableStore<Data>
+
+export function onSet<Data>(
+  store: Store<Data>,
+  handler: (payload: {
```
ai, line 7: What do you think about removing `payload` key? For the reason of shortness and byte saving =^_^=

lifecycle/index.d.ts
```diff
@@ -0,0 +1,35 @@
...
+    shared: any
+    stop(): void
+    abort(): void
+  }) => void
+)
+
+export function onChange<Data>(
+  store: Store<Data>,
+  handler: (payload: {
+    original: unknown[]
```
ai, line 18: What should be here? How we can avoid `unknown`?

map/index.d.ts
```diff
@@ -97,6 +92,17 @@ export interface MapStore<Value extends object = any> {
...
   notify(key: AllKeys<Value>): void
+
+  /**
+   * Get store value.
+   *
+   * ```js
+   * store.get()
+   * ```
+   *
+   * @param value Store value.
```
ai, line 103:
```suggestion
   * @return Store value.
```

atom/index.d.ts
```diff
@@ -53,6 +40,17 @@ export interface ReadableStore<Value = any> {
...
   listen(listener: (value: ReadonlyIfObject<Value>) => void): () => void
+
+  /**
+   * Get store value.
+   *
+   * ```js
+   * store.get()
+   * ```
+   *
+   * @param value Store value.
```
ai, line 51:
```suggestion
   * @returns Store value.
```

lifecycle/index.js
```diff
@@ -0,0 +1,76 @@
...
+    if (!current.length) {
+      storeContainer.delete(eventKey)
+      storeContainer.get(`${eventKey}-d`)()
+      storeContainer.delete(`${eventKey}-d`)
+    }
+    if (!storeContainer.size) container.delete(store)
+  }
+}
+
+export const onCreate = (destStore, cb) =>
```
ai, line 32: Can we rename it to `onStart` (to rhythm with `onStop`)? We may need `onCreate` to run listener on store creating from store template.

atom/index.js
```diff
@@ -1,64 +1,49 @@
...
-        if (!nextListeners.length) {
-          setTimeout(() => {
-            if (store.active && !nextListeners.length) {
-              if (destroy) destroy()
-              destroy = undefined
-              store.active = undefined
-            }
-          }, STORE_CLEAN_DELAY)
-        }
+        store.lc--
```
vadirn, line 34: Hey 👋 Looks like this might produce incorrect results if unsubscribe function is called multiple times

ai: Oh 😓 Can you show some example?

vadirn: For example `atom/index.test.ts`

```typescript
it('listen', () => {
  expect.assertions(4)
  let store = atom({ some: { path: 0 } })
  let unbind = store.listen(value => {
    expect(value).toBe(store.get())
  })

  store.set({ some: { path: 1 } })
  store.set({ some: { path: 2 } })
  expect(store.get()).toEqual({ some: { path: 2 } })
  unbind()
  unbind()
  expect(store.lc).toEqual(-1)
})
```
would pass, meaning `store.lc` checks might not work as expected. I am not sure how bad this can be. Just something I stumbled upon when trying to extend Svelte's stores with some extra optimisations.

vadirn: In my case subscribers are kept in a Set, so I just check its size instead of keeping a separate counter

ai: Thanks. Fixed cc4d514]

I tried `Set` for listeners, but it leads to drop of 25% [benchmark](https://github.com/nanostores/nanostores/blob/next/test/benchmark.js) results

PR #63, opened 2021-09-28
add note about nextjs11.1 workaround

Mitsunee:
NextJS 11.1 added a flag to enable their (currently experimetal) ESM support. This option is available as a preview for NextJS 12, which will support ESM out of the box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -619,4 +619,7 @@ module.exports = withTM({
 })
 ```
 
+In Next.js 11.1 you can alternatively use the experimental [`esmExternals`] config option.
+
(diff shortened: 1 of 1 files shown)


ai:
Should we just remove old Next.js instructions?

Mitsunee:
I think it's better to keep them for now in case someone is using older versions of next or prefers the plugin method

ai:
OK. I will remove old method a few months later.

Mitsunee:
just noticed the brackets around the url at the bottom of the file made the link not show up... learned yet another thing today.

PR #64, opened 2021-10-02
Lifecycle cleanup

ai:
1. Replace global listeners `Map` to `store.events` to avoid memory leaks probability.
2. Rename common words in arguments to explain how they are used
3. Refactor types (need to be finished)


mount/index.js
```diff
@@ -1,26 +1,24 @@
...
-      setTimeout(() => {
-        if (store.active && !store.lc) {
-          destroy && destroy()
-          destroy = undefined
-          store.active = undefined
-        }
-      }, STORE_CLEAN_DELAY)
-    })
-  ]
+  let unbinStart = onStart(store, () => {
```
Eddort, line 8:
```suggestion
  let unbindStart = onStart(store, () => {
```

mount/index.js
```diff
@@ -30,8 +28,7 @@ export const mount = (store, cb) => {
     }
   }
   return () => {
-    for (let unsub of unsubs) {
-      unsub()
-    }
+    unbinStart()
```
Eddort, line 31:
```suggestion
    unbindStart()
```

lifecycle/index.d.ts
```diff
@@ -1,35 +1,91 @@
...
+export function onSet<Value, Shared = never>(
+  store: Store<Value>,
+  listener: (payload: { args: [Value]; shared: Shared; abort(): void }) => void
 )
 
+/**
+ * Add listener to notifing about store changes.
+ *
+ * ```js
+ * import { onSet } from 'nanostores'
```
Eddort, line 29: It's needed to correct functions names in all JSDoc's

PR #66, opened 2021-10-02
Nanostores actions

Eddort:


action/index.d.ts
```diff
@@ -0,0 +1,24 @@
...
+ * setName('John')
+ * ```
+ * @param store Store instance.
+ * @param actionName Loging name.
+ * @param cb Your own function.
+ * @returns Your own function.
+ */
+export function action<
+  Store extends WritableAtom,
+  ActionName extends string,
```
ai, line 16: We can use `store: WritableAtom, actionName: string` below.

We need to create this type templates for the cases where two arguments share the single type like:

```ts
<Objects extends object, Key extends keyof Object>(store: Object, key: Key)
```

action/index.d.ts
```diff
@@ -0,0 +1,24 @@
...
+  Callback
+>(store: Store, actionName: ActionName, cb: Callback): Callback
```
ai, line 18: `cb` type should be something like `(store: Store) => void` (here we need type template for `Store`.

action/index.d.ts
```diff
@@ -0,0 +1,24 @@
...
+ * ```
+ * @param store Store instance.
+ * @param actionName Loging name.
+ * @param cb Your own function.
+ * @returns Your own function.
+ */
+export function action<
+  Store extends WritableAtom,
+  ActionName extends string,
+  Callback
```
ai, line 17:
```suggestion
  Callback extends (...args: any) => any
```

action/index.d.ts
```diff
@@ -13,12 +13,10 @@ import { WritableAtom, MapStore } from '../index.js'
  */
 export function action<
   Store extends WritableAtom,
```
ai, line 15: You do not need `Store` type template

PR #82, opened 2021-11-06
feat(lifecycle): `onAction` instead of `onError`

euaaaio:


ai:
Can I ask you to resolve the conflicts?

@Eddort what do you think about new API?

euaaaio:
> @euaaaio Can I ask you to resolve the conflicts?

Done

Eddort:
I think we need to add test cases to async action. The current realization doesn’t handle async actions. I will do PR later.

euaaaio:
> I think we need to add test cases to async action. The current realization doesn’t handle async actions. I will do PR later.

What do you mean? `onAction` is only about async actions.

Eddort:
When we call two same async actions we can't understand where are correct changes are. It is a race of promises.
I did the test case for it: https://github.com/euaaaio/nanostores/pull/1/files#diff-79c620f45d8436c15063845e42cba347a913694d2baa45263f0e2b7f4cae93e6R303

lifecycle/index.js
```diff
@@ -156,11 +156,42 @@ export let onMount = (store, initialize) => {
...
+
+export let onAction = (store, listener) =>
+  on(store, listener, ACTION, runListeners => {
+    let errorListeners = {}
+    let endListeners = {}
+    let originAction = store.action
     let originError = store.error
-    store.error = (error, actionName) => runListeners({ error, actionName })
+    let originEnd = store.end
+    store.action = (id, actionName, args) => {
```
ai, line 165: Adding methods to `store` can lead to dangerous things. User extends store object with custom methods/properties.

But it is just a thinking out loud. Anyway, we already started it and should fix it in another PR.

PR #86, opened 2021-11-29
Add type for notify function

AleksandrSl:
- [x] Do we allow calling notify without parameter? (See example with function as value)
- [x] Is it for writable stores only?

### DIFF
--- a/atom/errors.ts
+++ b/atom/errors.ts
@@ -7,9 +7,16 @@ store.listen(value => {
   value.value = 2
 })
 
+store.notify()
+store.notify("value")
+// THROWS Argument of type '"nonExistentKey"' is not assignable to parameter of type '"value" | undefined'.
+store.notify("nonExistentKey")
+
 let fnStore = atom<() => void>(() => {
   fnStore.set(() => {})
 })
 
 let fn = fnStore.get()
 fn()
+
+fnStore.notify()
--- a/atom/index.d.ts
+++ b/atom/index.d.ts
@@ -1,5 +1,7 @@
 import { lastAction } from '../action/index.js'
 
+type AllKeys<T> = T extends any ? keyof T : never
+
 type ReadonlyIfObject<Value> = Value extends undefined
   ? Value
   : Value extends (...args: any) => any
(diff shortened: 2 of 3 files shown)


atom/index.d.ts
```diff
@@ -1,5 +1,7 @@
 import { lastAction } from '../action/index.js'
 
+type AllKeys<T> = T extends any ? keyof T : never
```
AleksandrSl, line 3: Moved here because we need it here, and `atom` is imported in `map`, where this definition was before

atom/errors.ts
```diff
@@ -7,9 +7,16 @@ store.listen(value => {
...
+store.notify("nonExistentKey")
+
 let fnStore = atom<() => void>(() => {
   fnStore.set(() => {})
 })
 
 let fn = fnStore.get()
 fn()
+
+fnStore.notify()
```
AleksandrSl, line 22: Causes ts error for now? Should it be so?

ai: It means that all keys was changes. `changedKey` could be `undefined`.

AleksandrSl:
The second question still holds, do we allow notify in writeable only? Because we can't change anything

AleksandrSl:
If we have another node/group under the position rendering pipeline will shift one of these nodes to be stacked with a little shift. This mechanism is UI only, the real position of the element is the same. This leads to some quirks in the movement of these shifted nodes.
I think there is nothing to stop us from changing the real position of the node. At least we are talking about adding/pasting. We could find free space before creating the node.

I know that this solution doesn't cover all cases. The old mechanism should still exist, because of multiple editing. Two users add s node in the same place, a free position will be identical, nodes will end up in the same place. But for the time being, it's a more simple fix than removing movement quirks and will help in most cases.

ai:
> The second question still holds, do we allow notify in writeable only? Because we can't change anything

I do not see any reasons to allow notify for readable.

But we can do it if it will simplify code or make some edge cases simpler.

ai:
Thanks. Released in 0.5.7.

PR #87, opened 2021-12-09
types(map-template): export `AnySyncTemplate `

euaaaio:
Since we uses it in types of devtools and logger


ai:
Thanks. Released in 0.5.8.

PR #88, opened 2021-12-20
fix: 🔄 PropertyKey => KeyofBase

Beraliv:
* The bug that we found in ts-essentials – https://github.com/krzkaczor/ts-essentials/issues/278#issuecomment-998235314
* The explanation in the release – https://devblogs.microsoft.com/typescript/announcing-typescript-2-9-2/#support-for-symbols-and-numeric-literals-in-keyof-and-mapped-object-types

That's the way how to tolerate `keyofStringsOnly` option in TypeScript

### DIFF
--- a/map/index.d.ts
+++ b/map/index.d.ts
@@ -5,7 +5,9 @@ import type {
   AllKeys
 } from '../atom/index.js'
 
-type Get<T, K extends PropertyKey> = Extract<T, { [K1 in K]: any }>[K]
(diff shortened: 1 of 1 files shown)


ai:
Do we need to release it or the problem is theoretical?

Beraliv:
The problem is only in the project that uses `keyofStringsOnly`

If you didn't see the report yet, it means that noone used it yet

Issue #100, opened 2022-02-27
Computed stores fire in strange order making state incorrect


dkzlv:
[Here's a repro](https://codesandbox.io/s/pensive-williams-m72ikc?file=/src/App.js).

### Given

Atom 1 -> Computed 1 -> Computed 2

### Expected

When making changes to Atom 1, you always get the following order of execution: Computed 1 -> Computed 2.
Also, Computed 2 always has correct derived state.

### Actual

If you make changes in some sequence (I don't really get the logic), the order of execution of computed stores reverts, so first we get a Computed 2 execution with an OLD state (which results in incorrect derived state for this store), then Computed 1 gets executed, but it doesn't trigger Computed 2 execution.

Here's a video for this repro. Hint: if you see a string getting logged twice in a row, it's the bug. I did two doubleclicks, that should have resulted in number 3 and number 5 appearing.

https://user-images.githubusercontent.com/2846746/155877236-e5151f30-0375-43d2-9dcf-9183760b5002.mov

dkzlv:
Also, if you add `keepMount` to all stores, it all seems to be working fine.

ai:
Can you try 0.5.11? @btakita fixed similar bug.

dkzlv:
@ai I'm afraid it got even worse. Not only did it not fix the bug, now for some reason computed stores do not work at all—[even in the most basic cases](https://codesandbox.io/s/hopeful-swirles-3sxy85?file=/src/App.js).

I updated the initial repro to use latest version.

ai:
@btakita seems like there is regression in the fix

btakita:
@ai I think it was related to subsequent notifications occurring during a `notifyId` run. I added logic to compare the stores argument on the computed, running the `cb` when there's a change.

I was not able to reproduce the example from @dkzlv in a test or on my local machine. However, I had a similar regression & added a test which created the issue described above.

ai:
@dkzlv can you test 0.5.12?

dkzlv:
@ai My case seems to be working fine, issue is resolved. Thanks @btakita!

PR #103, opened 2022-03-12
fix: computed: notifies when stores change within the same notifyId

btakita:
fixes: https://github.com/nanostores/nanostores/issues/100

fix: index.d.ts: notifyId declaration


ai:
Thanks. Released in 0.5.12.

PR #114, opened 2022-07-11
Support `shared` object for `onMount`, Write more tests and Update docs

### LINKED ISSUE #113: `onStart` is called twice when `useStore` is used
https://codesandbox.io/s/kind-khorana-sced5q

At the example above:
* When using `onMount`, only `onMount` is printed to the console
* When using `onStart` together with `onStop`, `onStart onStop onStart` is printed to the console

Maybe I understand something incorrectly, but from my point of view:
**Expected behaviour:** pair `onStart` and `onStop` behave exactly like `onMount` with destructor
**Actual behaviour:** for some reason, `onStop` is called after the first `onStart`, and then `onStart` is called once again

### Some more information

According to these [tests](https://github.com/nanostores/nanostores/blob/main/lifecycle/index.test.ts#L18), `onStart` itself functions correctly.

Then I tried to test `onStart` function while using `useStore`, and the following test **fails**:

```ts
test('does onStart only once', async () => {
  let events = [];
  let simple = atom<string>('simple-value')

  onStart(simple, () => {
    events.push('start');
  })

  let TestA: FC = () => {
    let simpleValue = useStore(simple)
    return h('div', { 'data-testid': 'test' }, `1 ${simpleValue}`)
  }

  render(h(TestA))
  equal(events.length, 1);  // fail: got 2
})
```

**Conclusion:** the problem is either in my misunderstanding or in the combination of `onStart` and `useStore` or in some of them used separately.

Nagellan:
Fixes #113

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,7 +420,9 @@ Each store has a few events, which you listen:
 
 * `onStart(store, cb)`: first listener was subscribed.
 * `onStop(store, cb)`: last listener was unsubscribed.
-* `onMount(store, cb)`: shortcut to use both `onStart` and `onStop`. We recommend to always use `onMount` instead of `onStart + onStop`, because it has a short delay to prevent undesirable behavior.
+* `onMount(store, cb)`: shortcut to use both `onStart` and `onStop`. \
+  We recommend to always use `onMount` instead of `onStart + onStop`, \
+  because it has a short delay to prevent undesirable behavior.
 * `onSet(store, cb)`: before applying any changes to the store.
 * `onNotify(store, cb)`: before notifying store’s listeners about changes.
 
--- a/README.md
+++ b/README.md
@@ -420,7 +420,7 @@ Each store has a few events, which you listen:
 
 * `onStart(store, cb)`: first listener was subscribed.
 * `onStop(store, cb)`: last listener was unsubscribed.
-* `onMount(store, cb)`: shortcut to use both `onStart` and `onStop`.
+* `onMount(store, cb)`: shortcut to use both `onStart` and `onStop`. We recommend to always use `onMount` instead of `onStart + onStop`, because it has a short delay to prevent undesirable behavior.
 * `onSet(store, cb)`: before applying any changes to the store.
 * `onNotify(store, cb)`: before notifying store’s listeners about changes.
 
@@ -437,7 +437,7 @@ onSet(store, ({ newValue, abort }) => {
 })
 ```
 
-Event listeners can communicate with `payload.shared` object.
+Same event listeners can communicate with `payload.shared` object.
 
 
 ## Integration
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -420,7 +420,7 @@ Each store has a few events, which you listen:
 
 * `onStart(store, cb)`: first listener was subscribed.
 * `onStop(store, cb)`: last listener was unsubscribed.
-* `onMount(store, cb)`: shortcut to use both `onStart` and `onStop`.
+* `onMount(store, cb)`: shortcut to use both `onStart` and `onStop`. We recommend to always use `onMount` instead of `onStart + onStop`, because it has a short delay to prevent undesirable behavior.
```
ai, line 423: Can you keep 80 columns? I am old fashion :D.

README.md
```diff
@@ -437,7 +437,7 @@ onSet(store, ({ newValue, abort }) => {
 })
 ```
 
-Event listeners can communicate with `payload.shared` object.
+Same event listeners can communicate with `payload.shared` object.
```
Nagellan, line 440: To be honest, I expected the `shared` object to be available across different lifecycle methods.
Because of such confusion, I added this kind of info to the README.

ai: Is it hard to add `shared` for everyone? Will it affect performance?

Nagellan: > Is it hard to add `shared` for everyone? Will it affect performance?

Idk for now, I need time to deal with the codebase.
However, it seems like a topic for a separate PR

ai: > However, it seems like a topic for a separate PR

Totally agree

README.md
```diff
@@ -420,7 +420,9 @@ Each store has a few events, which you listen:
 
 * `onStart(store, cb)`: first listener was subscribed.
 * `onStop(store, cb)`: last listener was unsubscribed.
-* `onMount(store, cb)`: shortcut to use both `onStart` and `onStop`.
+* `onMount(store, cb)`: shortcut to use both `onStart` and `onStop`. \
```
ai, line 423:
```suggestion
* `onMount(store, cb)`: shortcut to use both `onStart` and `onStop`.
```

and the same on the line below

Nagellan: Strange
In VSCode line breaks don't work without the backslash:

image

ai: Hm. Could be VSCode bug.

After https://github.com/nanostores/nanostores/commit/0c34bad1599bf64715a73dc84dc9a00fbff0320c it looks good https://github.com/nanostores/nanostores#store-events

Nagellan: @ai not only VSCode but also GitHub =)

image

ai: Looks good. What is the problem?

Nagellan: I though you wanted to see 80 col width especially in a view mode.
For now I got you 😅

PR #115, opened 2022-07-15
Improve jsdocs, Fix onMount arg type

Nagellan:


lifecycle/index.d.ts
```diff
@@ -87,12 +87,11 @@ export function onNotify<Shared = never, SomeStore extends Store = Store>(
 /**
  * Add listener on first store listener.
  *
+ * We recommend to always use `onMount` instead.
  * See {@link onMount} to add constructor and destructor for the store.
```
Nagellan, line 91: I haven't noticed this line at the previous PR 😅
Now looks better

Nagellan:
Ok, let's wait until https://github.com/lukeed/tsm/issues/36 is solved.
If maintainer doesn't react, we can just specify NodeJS 18.5.0 version in GitHub workflow config or just make **Test / Node.js Latest Full (pull_request)** check optional, because it tests the code on an **unstable** version of NodeJS which is not cool for a required check imho.

ai:
We do not need to have green CI to merge docs PR

Nagellan:
@ai I updated the version of tsm, so now all checks are green

PR #117, opened 2022-07-27
Making an atom without an initial value should return possibly undefined

### LINKED ISSUE #116: Type safety issue with undefined store value
In TypeScript, the initial value of an atom is typed as optional. This means the following would typecheck, but fail with a runtime error:
```ts
import { atom } from "nanostores"
import { useStore } from "@nanostores/react"

const message = atom<string>()

export default function App() {
  const currentMessage = useStore(message)
  const firstWord = currentMessage.split(" ")[0]
  return <p>{firstWord}</p>
}
```
[Sandbox link](https://codesandbox.io/s/practical-galileo-zt17m5?file=/src/App.tsx:0-266)

I would suggest either...

- make the initialValue argument required
- add an overload which returns `WritableAtom<T | undefined>`

...although both could be breaking changes for TS users :sweat_smile:

itsMapleLeaf:
Closes #116

### DIFF
--- a/atom/errors.ts
+++ b/atom/errors.ts
@@ -1,4 +1,5 @@
-import { atom } from '../index.js'
+import { atom, WritableAtom } from '../index.js'
+import { isEqual } from '../test/helpers'
 
 let store = atom<{ value: string }>({ value: '1' })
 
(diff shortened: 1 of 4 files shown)


ai:
Looks good. I am going to party now. I will try to merge it tomorrow or after tomorrow.

ai:
Can I just fix `pnpm test` or CI?

itsMapleLeaf:
Oh, I didn't notice those errors :upside_down_face: You can go ahead and fix them

itsMapleLeaf:
Never mind, they were fairly straightforward fixes :sweat_smile:

PR #136, opened 2023-01-06
Update Vue example to use script setup

mat2ja:
- Follow modern Vue conventions
- Remove indentations from <script> block
- Put script block before template


ai:
@euaaaio need your review

euaaaio:
Sorry for the delay.
We already can use `setup` syntax since it is standard for the [documentation of single-file components](https://vuejs.org/guide/scaling-up/sfc.html).
And it’s ok to put the `script` block above the `template` in this repo for consistency.

@mat2ja thank you for your PR. 
@ai let's merge.

Issue #135, opened 2022-12-25
Seems like diamond dependency problem is still unsolved


istarkov:
[Here](https://github.com/nanostores/nanostores/pull/58/files) block 
```
### Diamond Problem
To make stores simple and small, Nano Stores doesn’t solve “Diamond problem”.
```
was removed from the readme. 

But simple test
```ts
import { atom, computed } from 'nanostores';
const main = atom(0);
const comp_a = computed(main, (v) => v);
const comp_b = computed(comp_a, (v) => v);
const last = computed([comp_a, comp_b], (a, b) => `${a} ${b}`);
last.subscribe((v) => {
  console.log('LAST', v);
});
main.set(1);
```

gives output

```log
LAST 0 0
LAST 1 0
LAST 1 1
```
So seems like the problem still exists or probably I don't understand the problem well and it's something other.
It doesn't somehow affect me, just found that [diamond](https://github.com/nanostores/nanostores/blob/main/computed/index.js#L7-L20) code playing with sources of this beautiful library, and haven't got how it could solve something like the above.

ai:
Could be related with https://github.com/nanostores/nanostores/issues/134

I am thinking of going into it later.

codeninja-ru:
so I submitted a PR for this #144 

the test case is

```javascript
test('prevents extra calls in a diamond', () => {
  const main = atom(0)
  const comp_a = computed(main, (v) => v)
  const comp_b = computed(comp_a, (v) => v)
  const last = computed([comp_a, comp_b], (a, b) => `${a} ${b}`)
  const events = []
  last.subscribe((v) => {
    events.push('LAST ' + v);
  });
  main.set(1)
  clock.runAll()

  equal(events, [
    'LAST 0 0',
    'LAST 1 1',
  ])

})

```

Stas-tolpekin:
Looks like to achieve the goal, it’s necessary to have an event channel to pass a message from parent store to children. Having this, the algorithm of setting the state can be the following:

1. The store sets the new value and passes the value to all children. (Children perform their logic upon of a parent state change, but do not call `listen` and `subscribe` callbacks)
2. The store send an event to all children to signal that all stores in the chain handled the new value. (On getting the event children call `listen` and `subscribe` callbacks)

This algorithm can be performed synchronously (without setTimeout). There could be applied additional optimizations (for instance, postpone calculation of value before getting the event if there are no any children). But looks like this approach can turn this simple store to pretty heavy construction.

[...]

PR #144, opened 2023-01-28
skipping extra computation calls (for diamond problem)

codeninja-ru:
In an example like this
`var a = computed([a, b], (a, b) => a + b);`
2 computation calls are possible
1. when `a` has changed
2. when `b` has changed
in diamond structure, both of the calls can be called altogether

In the proposed change, the computation is deferred with `setTimeout(..., 0)` and then the computation will be triggered only once. It may solve the diamond issue in a simple scenario. #135 (I added the test case this issue) and also `clock.runAll()` is need to be called in some tests

I believe in the current reactive archetype of nanostores completely solving of the diamond problem is not possible. Maybe it's not worth it at all.

### DIFF
--- a/computed/index.js
+++ b/computed/index.js
@@ -1,6 +1,8 @@
 import { onMount } from '../lifecycle/index.js'
 import { atom, notifyId } from '../atom/index.js'
 
+let callStack = [];
+
 export let computed = (stores, cb) => {
   if (!Array.isArray(stores)) stores = [stores]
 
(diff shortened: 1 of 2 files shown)


ai:
Adding `setTimeout()` will make it very hard to write tests and real code.

Diamond problem is less problem that having `await delay(1)` after every step in tests.

codeninja-ru:
It’s possible to replicate `setTimeout` without calling it. The same way as `clock.runAll()` does.

But the key problem is how to detect the end of all the queues. It can be done with introducing event system in atoms.

let's say
1. all `atom()` will send `onStart` and `onFinsh` events. For example `atom/index.js`
```javascript
        fire('onStart')
        for (let i = 0; i < listenerQueue.length; i += 3) {
          listenerQueue[i](listenerQueue[i + 1], listenerQueue[i + 2])
        }
        fire('onFinish')
```
2. computed atoms have two options
2.1. If it's a single value store, it will simply resend messages from the parent store
2.2.1 if it's a multi value store, it will send `onStart` on the first message from the parent and send `onFinish` on the last.
2.2.2 instead of calling `setTimeout` in `store.listen()` it will postpone the computation and start it on the last `onFinish` message 

Note: errors in store functions are needed to be handled since it may break the sequence

ai:
I decided to go this way https://github.com/nanostores/nanostores/pull/145

PR #145, opened 2023-01-29
Introduce level property to allow sorting of stores in computed

ogogorev:
Let's consider following example: 

```
let a = atom(0)
let b = computed(a, v => v)
let c = computed(b, v => v)
let d = computed([b, c], (b, c) => `${b} ${c}`)

d.subscribe((v) => {
  console.log('d', v);
});
```

The example above has following structure: `A -> B, B -> C, BC -> D`

It turns out that the problem is caused by the order in which listeners are added to the `b` store. The order of listeners follows the order of stores, passed to `d`. Since in the array `[b, c]` `b` comes first, `d` subscribes to `b`, then to `c`, and only then `c` subscribes to `b`. So `b` will notify `d` before `c`. Passing `[c, b]` solves the problem.

Letting users to manage the order of stores might not be the best option, since the relations between stores can be tricky and will cause bugs.

In my proposed solution I added to `atom` a new property `level`. Using `level` the computed store can sort passed stores in descending order and subscribe to them in the right order.

I also extended the test for the diamond problem, now it handles both `A -> B, A -> C, BC -> D` and `A -> B, B -> C, BC -> D` cases.

Intuition tells me that there might be more elegant solutions for the problem, especially using `notifyId` and `listenerQueue`.  I hope at least my explanation of the issue will be useful.

### DIFF
--- a/computed/index.test.ts
+++ b/computed/index.test.ts
@@ -68,18 +68,22 @@ test('prevents diamond dependency problem', () => {
   let store = atom<number>(0)
   let values: string[] = []
 
-  let a = computed(store, count => `a${count}`)
-  let b = computed(store, count => `b${count}`)
-  let combined = computed([a, b], (first, second) => `${first}${second}`)
+  let a = computed(store, count => count)
+  let b = computed(store, count => count)
+  let c = computed(b, count => count)
+  let combined = computed(
+    [a, b, c],
+    (first, second, third) => `${first}${second}${third}`
+  )
 
   let unsubscribe = combined.subscribe(v => {
     values.push(v)
   })
 
-  equal(values, ['a0b0'])
+  equal(values, ['000'])
 
   store.set(1)
-  equal(values, ['a0b0', 'a1b1'])
+  equal(values, ['000', '111'])
 
   unsubscribe()
 })
--- a/computed/index.test.ts
+++ b/computed/index.test.ts
@@ -2,7 +2,13 @@ import FakeTimers, { InstalledClock } from '@sinonjs/fake-timers'
 import { equal, ok } from 'uvu/assert'
 import { test } from 'uvu'
 
-import { StoreValue, computed, onMount, atom, STORE_UNMOUNT_DELAY } from '../index.js'
+import {
+  StoreValue,
+  computed,
+  onMount,
+  atom,
+  STORE_UNMOUNT_DELAY
+} from '../index.js'
 
 let clock: InstalledClock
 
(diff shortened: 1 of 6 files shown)


computed/index.test.ts
```diff
@@ -68,18 +68,22 @@ test('prevents diamond dependency problem', () => {
   let store = atom<number>(0)
   let values: string[] = []
 
-  let a = computed(store, count => `a${count}`)
-  let b = computed(store, count => `b${count}`)
-  let combined = computed([a, b], (first, second) => `${first}${second}`)
+  let a = computed(store, count => count)
```
ai, line 71: Can you keep old test logic with store names?

computed/index.test.ts
```diff
@@ -68,18 +68,22 @@ test('prevents diamond dependency problem', () => {
...
+    [a, b, c],
+    (first, second, third) => `${first}${second}${third}`
+  )
 
   let unsubscribe = combined.subscribe(v => {
     values.push(v)
   })
 
-  equal(values, ['a0b0'])
+  equal(values, ['000'])
```
ai, line 83: Can you keep `a0b0c0` thing here?

ogogorev: Now I see why the letters are important: turned out there was a bug with the order of stores in callback. Thanks!

ai:
I am right now in the travel and will think about it after tomorrow. Can you add more complex tests for now to be sure for edge cases?

ogogorev:
> I am right now in the travel and will think about it after tomorrow. Can you add more complex tests for now to be sure for edge cases?

Sure, will do it tomorrow.

ogogorev:
Well, I guess I was wrong saying that the problem is caused by the order of stores passed to a computed.
I will investigate it further and share all my findings.

ogogorev:
@ai
I found out that following code

```
let store = atom(0)

store.listen(v => console.log(v))

store.set(1)
store.set(1)
```

will log `1` two times. Is it intentional? I couldn't find any tests for this behaviour and would need your input.
I would actually expect only one log. But maybe I'm missing something.

I'm asking because my current solution includes memoizing of listeners, so the code above will log `1` only once.

ogogorev:
@ai Take a look on it please. 

Currently the `level` property is also added to `currentListeners` array, then in `notify` calls of listeners, whose `levels` are more than on step away from current level, are put in the end of `listenersQueue`.

Now it much more complicated than it was before, so I'm not sure it's worth it. But I also added some more complicated test cases, and almost all of them  (2, 3, and 4) are failing, if I run them on the `main` branch.

computed/index.test.ts
```diff
@@ -64,26 +70,136 @@ test('works with single store', () => {
...
+  equal(values, ['a0b0c0d0'])
 
   store.set(1)
-  equal(values, ['a0b0', 'a1b1'])
+  equal(values, ['a0b0c0d0', 'a1b1c1d1'])
 
   unsubscribe()
 })
 
+test('prevents diamond dependency problem 4 (complex)', () => {
```
ogogorev, line 153: Should I add a scheme for the data flow is this test? I think without a scheme nobody can understand it.

ai:
Thanks.  Released in 0.7.2.

Green369-1 (approved):

PR #146, opened 2023-01-30
Idea to solve a diamond problem

ramil-k:
Not a final code, just to show an idea.

# What if derived atom does not trigger notify on update?

Method run returns notify method.

### Common behavior
It triggers all listeners, collects resulting functions into a set of children notify methods.

### Derived atoms
It returns the set of notify methods.

### Top level atoms
It trigers all notify methods, collecting results into the same set and triggering those.

## Run method for the computed atom
If the value was updated, it returns notify method.

### DIFF
--- a/atom/index.js
+++ b/atom/index.js
@@ -20,7 +20,7 @@ export let atom = initialValue => {
       }
       return store.value
     },
-    notify(changedKey) {
+    notify(changedKey = undefined, isDerived = false) {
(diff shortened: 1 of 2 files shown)


ai:
Tests are falling

ramil-k:
> Tests are falling

Yes, it's the draft PR I created to try communicate the idea before digging deeper into the project.

PR #148, opened 2023-02-02
Infinite loop in listenerQueue

ogogorev:


ai:
Are we ready to merge it?

package.json
```diff
@@ -68,7 +68,7 @@
       "deprecated",
       "**/*.test.*"
     ],
-    "lines": 100,
+    "lines": 85,
```
ai, line 71: We need to continue keeping 100% coverage :)

ogogorev: Yeah, my assumption is that the coverage is not 100% for a long time.
I changed the command for tests to `uvu -b -r tsm . '\\.test\\.(ts|js)$'`, and then one check has failed because of the coverage. Here: https://github.com/nanostores/nanostores/actions/runs/4077570017/jobs/7026754129

My assumption is that with the old command was something wrong. But maybe I am missing something.

ogogorev: I ran on `main` branch these two commands:

1. old way to run `uvu`: `c8 tsm node_modules/uvu/bin.js . '\\.test\\.(ts|js)$'` - ends successfully
2. new way: `c8 uvu -r tsm . '\\.test\\.(ts|js)$'` - fails because the coverage is only 84.56% 

So it seems to me that the coverage is not 100%. Just nobody has noticed it yet.

If my assumptions are correct, I would keep 85% and fix it in the next PRs.

ai: > So it seems to me that the coverage is not 100%. Just nobody has noticed it yet.

Very likely that `uvu -r tsm` just breaks `c8` line counting.

ai:
Thanks. Released in 0.7.3.

PR #152, opened 2023-02-04
Update the setKey type to allow pass undefined as value

hgenru:
Based on the setKey code, if you pass undefined as a value into it, the field is deleted. That's why we should allow undefined to be passed as a value in types.

https://github.com/nanostores/nanostores/blob/5b0949d27741a4f9c1390bbca7f2a3a73de54e69/map/index.js#L10

### DIFF
--- a/map/index.d.ts
+++ b/map/index.d.ts
@@ -94,7 +94,7 @@ export interface MapStore<Value extends object = any>
    */
   setKey<Key extends AllKeys<Value>>(
     key: Key,
-    value: Get<Value, Key> | Value[Key]
+    value: Get<Value, Key> | Value[Key] | undefined
   ): void
 }
 


hgenru:
By the way, deleting a key when passing undefined as a value is not very obvious behavior. Especially when store is initialized with undefined in one of the fields. In that case not only that field will be deleted, but also unnecessary notify() will happen.

ai:
You should pass `undefined` only if store allows key deletion. As a store user, you should explicitly set that keys can be deleted by types:

```ts
interface User {
  id: string
  email: string
  name: string | undefined
}

user = map<User>({ id, email })

user.setKey('name', undefined) // OK
user.setKey('id', undefined) // Should be not OK to not broke the app
```

PR #161, opened 2023-02-27
Adding deepMap implementation

dkzlv:
TLDR: MapStore that's friendly to nested attributes and arrays.

Main magic happens in our own naïve implementation of object-path getting/setting. It uses recursion to only change object reference (via `structuredClone`) for the object that has some changed keys. It only works with plain objects and arrays, not sets, maps, getters/setters and other stuff.

As for tests, I copied most of them from `map`, but adjusted them/added new for nested keys scenarios.

Yet to do:

- [ ] documentation
- [ ] update react bindings so that `opt.keys` is statically typed for `deepMap`


.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules/
 
 coverage/
+.vscode
```
ai, line 4: IDE configs should be in your local git ignore file

deep-map/index.d.ts
```diff
@@ -0,0 +1,21 @@
...
+export type DeepMapStore<T extends Record<string, unknown>> = Omit<
+  WritableAtom<T>,
+  'setKey' | 'listen' | 'subscribe'
+> & {
+  setKey: <K extends AllKeys<T>>(key: K, value: GetPath<T, K>) => void
+  listen: Listener<T>
+  subscribe: Listener<T>
+}
+
+export function deepMap<T extends Record<string, unknown>>(
```
ai, line 19: We need TypeDocs here

deep-map/path.d.ts
```diff
@@ -0,0 +1,21 @@
+import type { Call, Objects } from 'hotscript'
```
ai, line 1: We will need to add `hotscript` to project’s `dependencies`

deep-map/path.js
```diff
@@ -0,0 +1,72 @@
+/**
+ * Get a deep value by key. Undefined if key is missing.
+ * Doesn't have a lot of runtime sanity checks, instead relies on TS to tell the user about
+ * their mistakes.
+ *
+ * @param obj Any object you want to get a deep path of
+ * @param path Path splitted by dots. Arrays accessed the same as in JS: props.arr[1].nested
+ */
+export const getPath = (obj, path) => {
```
ai, line 9: We are using `export function X()` for these cases. Let’s keep consistency.

deep-map/index.d.ts
ai: I like the idea. Some users already asked for something like this.

dkzlv: @ai Have a look. Added docs to all exported functions and to README itself.

deep-map/index.d.ts
```diff
@@ -0,0 +1,65 @@
+import { WritableAtom } from '../atom/index.js'
+import { AllKeys, GetPath } from './path.js'
+
+export * from './path.js'
+
+type Listener<T extends Record<string, unknown>> = (
+  listener: (value: T, changedKey: undefined | AllKeys<T>) => void
+) => () => void
+
+export type DeepMapStore<T extends Record<string, unknown>> = Omit<
```
ai, line 10: Can we pass `deepMap` to argument for `atom` after omitting and replacing this methods?

deep-map/index.js
```diff
@@ -0,0 +1,15 @@
+import { atom } from '../atom/index.js'
+import { getPath, setPath } from './path.js'
+
+export * from './path.js'
```
ai, line 4: Can we move this function here? Otherwise this file become too small 😅

pnpm-lock.yaml
```diff
@@ -20,12 +20,16 @@ specifiers:
...
   eslint-plugin-promise: ^6.1.1
+  hotscript: ^1.0.8
   nanodelay: ^2.0.2
   size-limit: ^8.2.4
   tsm: ^2.3.0
   typescript: ^4.9.5
   uvu: ^0.5.6
 
+dependencies:
+  hotscript: 1.0.8
```
ai, line 31: Can we avoid dependency (for insurance by copying types helpers)?

Right now it is zero-dependency project which is one of the argument why it is minimalistic. I am ok to loose this feature for a big feature, but not here.

deep-map/index.js
dkzlv: Currently modules are separated by their function. I surely can merge them, but it'll both make type notations harder to read/edit and the module itself will be messier.

> file become too small

… so?

deep-map/index.d.ts
dkzlv: Can you explain it once again? I didn't follow.

pnpm-lock.yaml
ai: > What kind of insurance?

Can we avoid dependency (for *instance by copying types helpers)?

dkzlv: I mean, insurance from what exactly?

ai: No insurance, it was just typo.

Having extra dependency is loosing one of our feature. How many code do we use from dependency? Can we just copy it?

dkzlv: We use quite a lot of their code. These are just two type functions (`Objects.AllPaths` and `GetFromPath`), but those are tightly entangled with the rest of the lib, and for a good reason (making the same features without this entanglement is quite hard, I was talking about this in Slack).
If I were to copy everything those two function rely on I would copy at least 150-200 LOCs of code that I couldn't support in any way. And I can't really say if I can write those helpers myself in a meaningful time.

Still, I don't understand what you mean by "zero dependency". This is pure TS types, it won't bloat the bundle in any way. I would think "zero dependency" is mostly about code-emitting deps. We can rename the feature from "Zero dependencies" to "Zero runtime dependencies" 😂

dkzlv: On top of that, if we one day decide to move away from `.js` + `.d.ts` to plain `.ts`, `hotscript` would actually simplify most of the typings we have.

ai: > This is pure TS types, it won't bloat the bundle in any way

People will not try to understand what type of dependency do they see. They still will see a dependency and will think about the project that it is not so `nano-` anymore.

deep-map/index.d.ts
ai: Does this code work:

```ts
import { deepMap, ReadableAtom } from 'nanostore'

function someStoreFunc (store: ReadableAtom) {
  …
}

someStoreFunc(deepMap())
```

dkzlv: Yeah, sure, it will. `ReadableAtom` has a very simple signature.

PR #178, opened 2023-04-13
docs: add details to Vanilla JS section

adhrinae:
- The example code now exposes the `subscribe` method.
- Added comments about the `listen` method for clarity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -560,7 +560,7 @@ export class AppComponent {
 
 ### Vanilla JS
 
-`Store#subscribe()` calls callback immediately and subscribes to store changes.
+`Store#subscribe(cb)` calls callback immediately and subscribes to store changes.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -573,6 +573,11 @@ function render () {
   console.log(`${post.title} for ${profile.name}`)
 }
 
+profile.subscribe(render)
```
ai, line 576: Calling `subscribe()` has no reason for this case.

Maybe we can explain the difference on some more reasonable example with a single store?

adhrinae: > Maybe we can explain the difference on some more reasonable example with a single store?

I agree. That's better than an arbitrary example that can't be understood. What would that example be like?

TBH, It would be better if the doc could explain common methods like `subscribe`, `listen` etc.

PR #181, opened 2023-04-17
Check referential equality before notifying

TrySound:
React useSyncExternalStore triggers rerenders only when value is not referentially equal to previous one.

Here updated atom to not notify subscriptions when new value is same.

This change will let as avoid additional checks in user land and rely more on store.subscribe in non-react logic.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,4 @@
 {
-  "packageManager": "pnpm@8.2.0",
   "name": "nanostores",
   "version": "0.8.0",
   "description": "A tiny (334 bytes) state manager for React/Preact/Vue/Svelte with many atomic tree-shakable stores",
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "packageManager": "pnpm@8.2.0",
   "name": "nanostores",
   "version": "0.8.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "packageManager": "pnpm@8.2.0",
```
ai, line 2: pnpm has a release every week. It will be too hard for me to update version in every `package.json` on every release

TrySound: This is a way to use reproducible version of package manager when corepack is enabled. Like lockfile. Pnpm is quite stable so maybe no need to bump every week. It's not a dependency affecting end user.

TrySound: For example I don't have locally installed yarn and pnpm and only rely on corepack. This field is the only requirement.

ai: What will happen when I update `pnpm `to `8.2.1`? Do I need to manually go to every open source project and update `package.json`?

TrySound: I'd say only if lockfile of the project is changed. But if you use corepack then you don't update your global pnpm but update the project to use different version.

If too much burden I'll remove it.

ai: Let’s remove it. I understand that it is nice to have, but current implementation of `corepack` is just not useful for real cases.

TrySound: Done

ai:
Merged to next major release branch

PR #184, opened 2023-04-18
Avoid copying listeners

TrySound:
Found the source of change here https://github.com/nanostores/nanostores/commit/242d2b123ba4ccb7f11687fae40f70a3878b8bcc

Though back then listeners were called in the same loop. Now listeners are called when iterating over listenerQueue which makes copying unnecessary.


TrySound:
Btw for pr merges you can choose automatically better description instead of list of commits
image

PR #185, opened 2023-04-19
Drop notifyId and remove store.notify from public api

TrySound:
`store.notify()` was used to enforce update when no value is changed though this case is not valid in react which checks referential equality before update.

Here remove store.notify from types and dropped notifyId which was used to enforce recomputing of dependent computed stores.

### DIFF
--- a/atom/errors.ts
+++ b/atom/errors.ts
@@ -7,19 +7,12 @@ store.listen(value => {
   value.value = 2
 })
 
-store.notify()
-store.notify('value')
-// THROWS '"nonExistentKey"' is not assignable to parameter of type '"value"'
-store.notify('nonExistentKey')
-
 let fnStore = atom<() => void>(() => {
   fnStore.set(() => {})
 })
 
 let fn = fnStore.get()
 fn()
-
-fnStore.notify()
 
 let store2 = atom<string | undefined>()
 store2.set("new")
--- a/atom/index.d.ts
+++ b/atom/index.d.ts
@@ -87,19 +87,10 @@ export interface WritableAtom<Value = any> extends ReadableAtom<Value> {
    * @param newValue New store value.
    */
   set(newValue: Value): void
-
-  /**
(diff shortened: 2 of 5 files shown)


ai:
@btakita do you think that it is safe to remove `nottifyId` right now?

blikblum:
It should consider all frameworks, not only React.

In my case, notify is useful. I have a class that stores some internal data. When the internal data chances I call notify.

I know all benefits of imutability, but that are other valid ways of developing state.

ai:
I know that it is nice to have immutability, but it just doesn’t scale on the real API of frameworks. Many frameworks require mutability.

It applies not only to React, but, as I know, we have similar problem in Vue.js.

btakita:
`notifyId` & `diamondNotifyId` were added to ensure diamond dependencies are not notified multiple times. The diamond dependency tests should be sufficient to tell if `notifyId` is necessary or not.

TrySound:
Args check does the same job so now we would have one less state to store in module scope.

btakita:
I see. I took a better look at the PR. I also use mutable state for performance reasons. If we go ahead with this change, a mutable option or extension library with mutable computeds would be appreciated.

From https://github.com/feltcoop/svelte-mutable-store

> Sometimes mutation is required, like with the unclonable WeakMap and APIs that mutate things outside of your control, and other times it's desirable, like with large collections that are expensive to copy.

btakita:
Here is a pattern which could allow updates to immutable atoms with mutable objects.

```ts
const wm = new WeakMap()
const wm_atom_o = atom({ value: wm })
const key = {}
wm.set(key, 1)
wm_atom_o.set({ value: wm })
```

I think we can go ahead with this PR with some documentation on ways for mutable objects to notify the subscribers. I also support doing this as `notifyId` is not part of the public api & computed uses `notifyId`. This means an external library cannot reimplement computed the same same behavior.

PR #186, opened 2023-04-19
Fix README typo: .listen() -> .subscribe()

dsernst:
Simple docs typo fix: The example is supposed to be for `Store#subscribe()`, not `.listen()`.


ai:
Thanks!

PR #189, opened 2023-05-08
docs(action): update JsDocs example for "action"

Royserg:
JsDocs for `action` seems to be out of sync with the example provided in main README - first argument of callback is the `store`.


ai:
Thanks!

Issue #171, opened 2023-03-09
Context


ai:
Let’s discuss a way to solve a few problems

## Problem 1: Cache API

Many stores have cache:
* Logux Client caches resource’s instances
* Nano Stores Query caches responses

Current Nano Stores has `MapTemplate` for such things. But it has a few problems:
1. Can use only a string ID, which doesn’t fit Nano Stores Query needs.
2. People do not really understand it. There are many questions how to use `MapTemplate`.
3. The API is almost private. It is hard to use it.

## Problem 2: SSR

In SSR, Node.js will render HTML for multiple users in the same time (in the same Node.js environment). If you just export store from a file, all users will share the same store. It could be a very dangerous.

For instance, if we have something like `import { authStore } from '../store/auth.js'` will be the same store for all users.

We need a way to separate stores from user’s environment

## Problem 3: Mocks

It will be nice to be able to replace any store with a mock for tests or to create an environment for Storybook.

Right now we can use `atom#set()` or some stores provide custom API. But it doesn’t disable events, and it will be nice to have some common API.

## Problem 4: DevTools

It will be nice to have some sort of `Map` or `Set` with all stores on the current page, so we will be able to use them in DevTools.

geminigeek:
hi,
just  came across Problem 2, i am using it with astro + vue , for SSR using ( Astro node adaptor as middleware ), any solution for the issue ?

ai:
@geminigeek the best solutions is to export store creators and not stores themselves

```js
export function createProfileStore () {
  return atom()
}
```

And then create a single instance in application root.

ai:
My suggestion of Context API:

1. Nano Stores exports function to create a context: `import { createContext } from 'nanostores'`
2. The one global context will be created by default : `import { globalContext } from 'nanostores'`
3. All events (`'onMount`, etc) get current context to the callback.
4. Store’s access API for stores will have a 2 forms:
   1. Global context: `store.get()`, `store.listen()`
   2. User’s context: `withContext(store, context).get()`, `withContext(store, context).listen()`
5. Context has an API to:
   1. Get all registered stores
   2. Mock any store’s value without calling events
5. All libraries much use `withContext()`. In the app, developer could choose between global or user’s context (it will be illegal to mix global and user’s to avoid mistakes).
   6. Calling `createContext` will “poison” global context and any calls on global context will throw an error
7. React’s `useStore()` will look for Context from special React’s context.

TrySound:
[...]

PR #190, opened 2023-05-09
Store atom state in context

TrySound:
Ref https://github.com/nanostores/nanostores/issues/171

Here's initial attempt to move all stores state into global context object. Context is basically a listeners queue and a map of stores states.

Context management api is still yet to explore. For now it's for internal usage now.
The important thing here is accessing context with getter function which can allow to use AsyncLocalStorage in node and cloudflare workers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -550,8 +550,8 @@ It passes store’s value to callback.
 ```js
 import { profile } from '../stores/profile.js'
 
-profile.listen(() => {
-  console.log(`Hi, ${profile.name}`)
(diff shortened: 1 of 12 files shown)


ai:
Thanks. I plan to review it when I will come back from China, on next week.

ai:
At first glance it looks solid for me. But I am afraid to release it in 0.9 and will release your previous PRs first.

TrySound:
Sure, sounds fine

PR #192, opened 2023-05-16
Grammar correction

phocks:
"A unique feature of Nano Stores is that every state has two modes."

The corrections made include:

    - "Nano Stores" should be followed by "is" because it functions as a singular noun phrase.
    - "Every state" should be followed by "has" to match the singular form.

The corrected sentence conveys that Nano Stores has a unique feature where each state has two modes.


ai:
Thanks!

PR #193, opened 2023-05-17
add peek method

dartess:
Useful for solving this problem: https://github.com/nanostores/react/issues/17
Also help to fix this test: https://github.com/nanostores/i18n/pull/32/files#diff-ce9cd3e0f8d680b94d10ce764b503c910e4d03350c9e8256799087cb9e944eceR245

### DIFF
--- a/atom/index.js
+++ b/atom/index.js
@@ -21,6 +21,9 @@ export let atom = (initialValue, level) => {
       }
       return store.value
     },
+    peek() {
+      return store.value
+    },
     notify(changedKey) {
       currentListeners = nextListeners
       let runListenerQueue = !listenerQueue.length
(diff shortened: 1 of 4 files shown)


atom/index.js
```diff
@@ -21,6 +21,9 @@ export let atom = (initialValue, level) => {
       }
       return store.value
     },
+    peek() {
+      return store.value
```
ai, line 25: `store.value` is not always `Value`. Is store has no subscribers, the `store.value` will be `undefined` 

Here is an example
https://github.com/nanostores/nanostores/blob/main/computed/index.js#L20

I am thinking of moving `store.value` to a public API with `Value | undefined` type. Will it work?

ai:
I moved `Store#value` to readonly public instead d2fa758

PR #199, opened 2023-05-27
Store atom state in context

TrySound:
Ref https://github.com/nanostores/nanostores/issues/171

Here's initial attempt to move all stores state into global context object. Context is basically a listeners queue and a map of stores states.

Context management api is still yet to explore. For now it's for internal usage now.

### DIFF
--- a/atom/index.js
+++ b/atom/index.js
@@ -1,34 +1,38 @@
 import { clean } from '../clean-stores/index.js'
-
-let listenerQueue = []
+import { getListenerQueue, getStoreState } from '../context/index.js'
 
 export let atom = (initialValue, level) => {
-  let listeners = []
   let store = {
-    lc: 0,
     l: level || 0,
-    value: initialValue,
+    iv: initialValue,
+    get lc() {
+      let state = getStoreState(store)
+      return state.lc
+    },
+    get value() {
+      let state = getStoreState(store)
+      return state.v
+    },
     set(data) {
-      if (store.value !== data) {
-        store.value = data
+      let state = getStoreState(store)
+      if (state.v !== data) {
+        state.v = data
         store.notify()
       }
     },
     get() {
-      if (!store.lc) {
+      let state = getStoreState(store)
+      if (!state.lc) {
         store.listen(() => {})()
       }
-      return store.value
+      return state.v
     },
     notify(changedKey) {
+      let state = getStoreState(store)
+      let listenerQueue = getListenerQueue()
       let runListenerQueue = !listenerQueue.length
-      for (let i = 0; i < listeners.length; i += 2) {
-        listenerQueue.push(
-          listeners[i],
-          store.value,
-          changedKey,
-          listeners[i + 1]
-        )
+      for (let i = 0; i < state.ls.length; i += 2) {
+        listenerQueue.push(state.ls[i], state.v, changedKey, state.ls[i + 1])
       }
 
       if (runListenerQueue) {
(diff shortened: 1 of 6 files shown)


ai:
Here we have only a global context switcher. Will it cover all our cases?

TrySound:
SSR in node and cloudflare can be solved with async local storage. The main idea here is to always use getter. We can provide a way to inject storage getter and always get the right context.

ai:
1. In React, we will pass Nano Stores Context by React Context and use it in `useStore(store)`. What API we are planning to use `useStore()`?
2. Should we re-use the context in `computed()` in dependence stores?

TrySound:
1. Yes, we can rely on provided context in useStore as framework specific way. I thought of using synchronous wrapper `provide(context, () => store.get())`.
2. Wanted to cover non-atom cases in separate PR to simplify review.

ai:
> provide(context, () => store.get()).

How it will work?

> Wanted to cover non-atom cases in separate PR to simplify review.

Sure, here we need only a plan

TrySound:
Some global variable set right before invoking callback. It will be used by context getter. Since the code is synchronous the other invoke will override after tick is complete.

ai:
But store could have `setTimeout()`/`fetch()` callbacks, where we need to pass current context

TrySound:
Here we talking about useStore. For those I guess we need async local storage or explicitly provided context.

ai:
I see three steps:

1. Pass context to `listen`, `onMount`, and other events.
2. Get a simple API to use this context during the work with other store in callback
3. Add a syntax sugar with special async API for stores (where we can hide systems from steps 1 and 2).

But we need to create a whole plan before writing a code.

TrySound:
Listen is straightforward. Queue is stored in context too.

```
provide(context, () => {
  store.listen(() => {}0
})

provide(context, () => {
  store.set(value)
})
```

Lifecycles are basically overrides of builtin methods and can access context with getter too. So there is no need to specify context for them too.

ai:
> Listen is straightforward. Queue is stored in context too.

I am talking about:

```js
onMount($comments, async () => {
  let comments = await fetch('/comments')
  $comments.set(comments)
})
```

or:

```js
$form.listen(async form => {
  if (form.ready) {
    let result = await fetch('/comments')
    $form.setKey('result', result`)
  }
})
```

TrySound:
Yes, and here only 2 options synchronous provide in place or global async local storage.
It's not onMount or store.listen api as store.set is invoked directly.

One more option is do some context binding like `const boundStore = bindContextToStore(context, store)` but it will need to override all methods and gutters which is not very generic.

TrySound:
Here's how async local storage api can look like

```
import { AsyncLocalStorage } from 'node:async_hooks';
import { injectContextGetter } from 'nanostores';

const createRunner = () => {
  const storage = new AsyncLocalStorage();
  injectContextGetter(() => storage.getStore())
  return (callback) => {
    const context = createContext()
    storage.run(context, callback)
  }
}

const runWithContext = createRunner()

http.createServer((req, res) => {
  runWithContext(async () => {
    await renderApp();
  })
}).listen(8080)
```

blikblum:
As proposed in #171 , it would be nice to have a way to use the stores of a context in devtools. 


Now, in logger, is necessary to configure manually each store:

```js
import { logger } from '@nanostores/logger'

import { $profile, $users } from './stores/index.js'

let destroy = logger({
  'Profile': $profile,
  'Users': $users
})
```

To accomplish that, seems to be necessary a way to detect when a store is added to the context and a way to give the store a name.

Are you planning something like that?

ai:
Closing for https://github.com/nanostores/nanostores/pull/228

Issue #201, opened 2023-05-28
Computed behavior changed between 0.8.1 and 0.9.0


peschee:
Given this example:

```ts
const one = atom<string | undefined>(undefined);
const two = computed(one, (value: string | undefined) => !!value);
```

and this code:

```ts
console.log('one', one.get());
console.log('two', two.get());
```

in **`0.8.1`**, the output / values were

```
undefined
false
```

in **`0.9.0`**, the output / values are

```
undefined
undefined
```

Is this expected?

Example projects are here:

- 0.8.1: https://stackblitz.com/edit/vitejs-vite-jjawn6?file=src%2Fmain.ts
- 0.9.1: https://stackblitz.com/edit/vitejs-vite-m18s4j?file=src%2Fmain.ts

ai:
@TrySound could it be related with your changes?

TrySound:
Yes, notify always forced recomputing before. Now check fails when undefined is compared to out of range index. Need to add array length check additionally.

Good catch @peschee, thanks

Fixing

TrySound:
https://github.com/nanostores/nanostores/pull/202

ai:
The fix was released in 0.9.1

peschee:
Awesome, thanks for a quick fix!

PR #202, opened 2023-05-28
Fix computed with undefined argument

TrySound:
Ref https://github.com/nanostores/nanostores/issues/201

Notify always forced recomputing before. Now check fails when undefined is compared to out of range index.

Added existance check to diamond args to invalidate initial computation.


ai:
Thanks! Released in 0.9.1.

Issue #188, opened 2023-04-28
solidjs style computeds


btakita:
Solidjs signals use inline dependencies, removing the redundancy of declaring the dependency atoms when the computed is defined. I think we can lazily add dependencies when the computed function is executed.

Perhaps it could look something like:

```ts
const base = atom(0)
const dependency = computed(()=>base() + 1)
console.info({ dependency: dependency() })
dependency.unbind()
```

ai:
It looks interesting. Can somebody make an PR? We need to check the size of this feature.

btakita:
I can work on a PR when I have some free time, which will be at least 2 weeks from now. At the very least, this can be a separate package, but it may be small enough to include in this project. To anyone reading this who wants to work on this, don't let this message dissuade you from creating a patch.

btakita:
Hey @dtinth! I created [a PR](https://github.com/nanostores/nanostores/pull/203) which adds autosubscribe & async support. So your example would be supported as you wrote it using `computed` or since atoms are functions, can be called directly with autosubscribe in synchronous handlers.

```ts
const $shape = atom('circle')
const $radius = atom(4)
const $width = atom(2)
const $height = atom(3)

const $area = computed(() => {
  if ($shape() === 'circle') {
    return Math.PI * $radius() ** 2
  }
  if ($shape() === 'rectangle') {
    return $width() * $height()
  }
})
```

```ts
const $a = atom('a1')
const $b = atom('b1')
const $switcher = atom($a)
const $c = computed(() => {
  return $switcher()()
})
```

There is also support for async computeds. Re-entrancy is also supported, meaning if a callback is run before a prior callback completes, the prior callback will be marked as "stale". An example:

```ts
computed(async use => {
  let person = await fetch(`https://api.com/people/${personId()}`).then(response => response.json())
  if (use.stale()) return // stop run if the cb is stale
  let messages = await fetch(`https://api.com/people/${personId(use)}/messages`).then(response => response.json())
  return { person, messages } // This will not be saved if the cb is stale
})
```

btakita:
> I am impressed with the implementation of async. I think it would be useful, especially that you handled the "stale" case.

Thank you! There are some [additional use cases](https://github.com/btakita/nanostores/blob/issues/188/computed/index.d.ts#L17-L129) which includes **Event binding & unbinding** and **Helper functions & Private stores**.

The helper functions on `Autosubscribe` (`use`) are:

[...]

PR #203, opened 2023-06-04
`computed()` like in Solid.js & async `computed()`

btakita:
Implements https://github.com/nanostores/nanostores/issues/188

I implemented a version of computed which has 2 modes:

- predefined dependencies (which already exists in nanostares)
- inline dependencies (similar to solid-js signals)

The inline dependencies takes a `cb: (get: (atom: OriginStore)=>StoreValue<OriginStore>) => Value` argument defined:

https://github.com/btakita/nanostores/blob/issues/188/computed/index.d.ts#L35-L37

The `get` argument can be used `get(dependencyStore)` or `dependencyStore(get)`. Note that `dependencyStore(get)` has better type inference. I'm not sure how to define `get` to infer it's argument type, so any tips or patches are welcome.

Examples can be found in the [added tests](https://github.com/btakita/nanostores/blob/issues/188/computed/computedSignal.test.ts).

This has not fully optimized for size & I wanted to start the PR to begin discussing the best way to proceed.

There are a few commits. The earlier commits with [latest being 7e903e...](https://github.com/nanostores/nanostores/pull/203/commits/7e903e237fac7bf60a656c94609c5410321e0d81) implement a separate `computedSignal` function to implement the inline dependencies mode, leaving the predefined dependencies `computed` untouched.

In regards to size, the dual-mode `computed` is slightly larger at 1230 B vs 1100 B for the 2 function approach.

Note that both approaches have some room for size optimization.

Are either of these sizes acceptable to be included? Assuming that the dual-mode `computed` is slightly bigger, would it be better to have the dual-mode `computed` or should we opt to go with the 2 function approach which will be slightly smaller?


btakita:
I'm exploring the 2 function approach https://github.com/btakita/nanostores/tree/issues/188-2fn

In this branch I renamed `computedSignal` to `memo`, which is from solid-js's ontology. `memo` is a shorter & less noisy function name.

btakita:
I included memo in the size-limit & the results were different.

```json
{
  "size-limit": [
    {
      "name": "Atom",
      "import": {
        "./index.js": "{ atom }"
      },
      "limit": "389 B"
    },
    {
      "name": "No Memo All",
      "import": {
        "./index.js": "{ map, computed, action }"
      },
      "limit": "1100 B"
    },
    {
      "name": "All",
      "import": {
        "./index.js": "{ map, computed, action, memo }"
      },
      "limit": "1257 B"
    }
  ]
}
```

computed/computedSignal.test.ts
```diff
@@ -0,0 +1,362 @@
...
+  unbind()
+  clock.runAll()
+  equal(value, 'b 1')
+  equal(renders, 3)
+})
+
+test('works with single store', () => {
+  let number = atom<number>(1)
+  let decimal = computed(() => {
+    return number() * 10
```
TrySound, line 55: Do we want to change the api here? Maybe just use number.get()?

btakita: The solid-js signal/memo api uses pure functions. It is a cleaner syntax but it does add some size to the bundle. I'm personally fine with either. I have an extension library (@ctx-core/nanostores) which adds a Proxy function over atoms & computeds.

btakita: Oops, I didn't include the function api in the original issue. I'm ok removing the function api or leaving it in depending on what others think would be best.

btakita: I removed the atom function that aliases .get in https://github.com/btakita/nanostores/pull/new/issues/188-no-fn

The tests which include example usages can be compared at:

https://github.com/btakita/nanostores/blob/issues/188-no-fn/computed/computedSignal.test.ts
https://github.com/btakita/nanostores/blob/issues/188/computed/computedSignal.test.ts

btakita:
There are currently 3 options being investigated resulting in 8 different options in total. There are varying & minor package size implications for these options as well.

- 2 computed mode functions (`computed` & `memo`) vs a single dual-mode `computed` function
- an atom is a function which aliases the `.get` method vs keeping the same atom interface
- required `cx` callback function to be passed vs optional `cx` callback with a global stack to track synchronous computeds.

Note that these PR branches do not implement async `computeds`, but async support can be added by adding a `set` function argument.

## Code examples for the 4 permutations

### `computed` + `memo` & `atom` as a function & required `cx`

```ts
let firstName = atom('Joe')
let lastName = atom('Smith')
let fullName = memo(cx => `${firstName(cx)} ${lastName(cx)}`)
```

### `computed` + `memo` & `atom` as a function & optional `cx`

```ts
let firstName = atom('Joe')
let lastName = atom('Smith')
let fullName = memo(() => `${firstName()} ${lastName()}`)
```

### `computed` + `memo` & unchanged `atom` & required `cx`

```ts
let firstName = atom('Joe')
let lastName = atom('Smith')
let fullName = memo(cx => `${firstName.get(cx)} ${lastName.get(cx)}`)
```

### `computed` + `memo` & unchanged `atom` & optional `cx`

```ts
let firstName = atom('Joe')
let lastName = atom('Smith')
let fullName = memo(() => `${firstName.get()} ${lastName.get()}`)
```


### dual-mode `computed` & `atom` as a function & required `cx`

```ts
let firstName = atom('Joe')
let lastName = atom('Smith')
let fullName = computed(cx => `${firstName(cx)} ${lastName()}`)
```

### dual-mode `computed` & `atom` as a function & optional `cx`

```ts
let firstName = atom('Joe')
let lastName = atom('Smith')
let fullName = computed(() => `${firstName(cx)} ${lastName()}`)
```

### dual-mode `computed` & unchaged `atom` & required `cx`

```ts
let firstName = atom('Joe')
let lastName = atom('Smith')
let fullName = computed(cx => `${firstName.get(cx)} ${lastName.get(cx)}`)
```

### dual-mode `computed` & unchaged `atom` & optional `cx`

```ts
let firstName = atom('Joe')
let lastName = atom('Smith')
let fullName = computed(() => `${firstName.get()} ${lastName.get()}`)
```

The dual-mode `computed` & `atom` as a function & optional `cx` provides the "cleanest" api, but is adds the largest size increase for common usage. The `computed` + `memo` & atom` as a function & optional `cx` is slightly larger only if both `computed` + `memo` are imported.

btakita:
Following up on a couple of points:

## async support

If the `cb` in `computed(cb)` returns a PromiseLike, any code executed synchronously in the `cb` will be covered by implicit `cx`, which is pushed onto the global stack. After the microtask, the implicit `cx` will be popped off the global stack...which means after `await` in the `cb`, there is no `cx` on the global stack.

```ts
let $person_id = atom()
let $person_op_count = atom(0)
computed(
  async (cx, set) => {
    // cx is implicitly on the global stack
    if (!$person_id()) return set(null) // sets the value to null & returns a magic value (perhaps undefined) which is ignored by the computed api
    set({ loading: true }) // sets an intermediate loading value
    let person = await fetch(`https://www.example.com/people/${$person_id()}`).then(response => response.json())
    // cx is no longer implicit as it was popped off the global stack
    $person_op_count.set($person_op_count() + 1)
    return set(person)
    // or return can be used
    return person
  }
)```


## implicit `cx` vs explicit `cx`

Implicitly listening to an atom may not be what is intended & in some cases can lead to infinite recursion. In the example above, `$person_op_count.set($person_op_count())` was called. It would not cause infinite recursion with implicit `cx` only because it was called after the `await`. In the following example, infinite recursion would occur:

```ts
let $person_id = atom()
let $person_op_count = atom(0)
computed(
  async (cx, set) => {
    // cx is implicitly on the global stack
    if (!$person_id()) return set(null)
    set({ loading: true })
    $person_op_count.set($person_op_count() + 1) // CIRCULAR DEPENDENCY
    return fetch(`https://www.example.com/people/${$person_id()}`).then(response => response.json())
  }
)```

So an implicit `cx` is a cleaner api, however it does have the danger of unintended usage. This could be avoided by passing in a nullish value into the read, however that seems arcane.

```ts
let $person_id = atom()
let $person_op_count = atom(0)
computed(
  async (cx, set) => {
    // cx is implicitly on the global stack
    if (!$person_id()) return set(null)
    set({ loading: true })
    $person_op_count.set($p
erson_op_count(null) + 1) // No circular dependency
    return fetch(`https://www.example.com/people/${$person_id()}`).then(response => response.json())
  }
)```

So requiring `cx` to be passed into `.get` explicitly would not have these footguns. Most of the implicit `cx` use cases would not have the circular dependency issue & is a slimmer api, but would have this edge case.

**Comment by blikblum** [comment]
Just my 2c: is not this adding complexity and deviating from nanostores selling points of being tiny and simple?

**Comment by btakita** [comment]
> Just my 2c: is not this adding complexity and deviating from nanostores selling points of being tiny and simple?

Re: tiny Not sure. About 160 B would be added to the full package.

Re: simple, `computed` with many dependencies are complicated. Right now dependencies need to be referenced 3 times:

- in the `stores` dependency array
- an argument in the `cb` function
- in the body of the `cb` itself

Inlining the dependencies would require each dependency to be called in the `cb` body.

Here is a contrived example:

```ts
computed([dep0, dep1, dep2, dep3, dep4, dep5, dep6, dep7, dep8, dep9], (
  dep0, dep1, dep2, dep3, dep4, dep5, dep6, dep7, dep8, dep9
) => {
  if (dep0) return dep1
  else if (dep2) return dep 4
  else if (dep3) return dep 7
  else if (dep5) return dep 8
  else if (dep6) return dep 9
})
```

```ts
computed(() => {
  if (dep0()) return dep1()
  else if (dep2()) return dep 4()
  else if (dep3()) return dep 7()
  else if (dep5()) return dep 8()
  else if (dep6()) return dep 9k)
})
```

btakita:
There a couple more complexities with the current `computed` example above that would be avoided with inline dependencies.

## The $value/value naming issue

There is both a store object & store value that the programmer has to declare. The example above was actually not idiomatic. This is:

```ts
computed([$dep0, $dep1, $dep2, $dep3, $dep4, $dep5, $dep6, $dep7, $dep8, $dep9], (
  dep0, dep1, dep2, dep3, dep4, dep5, dep6, dep7, dep8, dep9
) => {
  if (dep0) return dep1
  else if (dep2) return dep4
  else if (dep3) return dep7
  else if (dep5) return dep8
  else if (dep6) return dep9
})
```

With the current `computed`, both $dep0 & dep0 need to be represented. The inline dependency `computed` only needs to declare `$dep0()`.

## Reconciling the `cb` argument order with the dependency order

In addition to 3 touch points per dependency, the order of the dependencies must match the order of the dependency array. It is possible to incorrectly reconcile the `cb` argument list with the dependency list. Inline dependencies does not have this issue.

Here is a current summary of the problems that inline dependencies solve:

- Avoids extraneous boilerplate ceremony code
- 1 touch point per dependency instead of 3 touch points per dependency
- No need to reconcile `cb` arguments with dependency list
- The 2 name $value/value issue is avoided
- 1 optional argument to the `cb` instead of n arguments to the `cb`
 
# It is worth the effort to put thought into this PR

It is worth the effort to scrutinize this PR to create the best possible api. Library apis have a tendency to stick, even if better solutions which are breaking changes are later found...due to the effort to update the dependent code bases. `computed` calls can be plentiful in a codebase, so a breaking change in the future would be costly.

## Improvements to this PR

I have considered some of the issues raised & how to avoid footguns, such as the circular dependency issue. I will propose some changes & then show an example with those changes.

### `atom` will be a function that can auto-listen

The `atom` will be a function that **can** auto-listen when called inside a `computed`. The existing props will be assigned to the `atom` function.

### `.get` will not auto-listen

This will avoid the circular-dependency issue in the example code above & to explicitly preserve the semantics of `.get` simply returning the current `value` as exists today.

### The inline `computed` `cb` will have 1 argument which is `cx`

A `cx` argument will be to support an async `computed` `cb`. A synchronous `computed` `cb` would not need to use `cx`.

A `cx` will be a function with a callback which will set `cx` onto the global stack during the synchronous execution of the callback. This will allow `atom` functions to auto-listen after an `await` is called in the `cb`.

`cx` will have a `.set` method to set intermediate values such as loading states or other state transitions.

`cx` will have a `.noop` prop which can be returned to prevent setting `computed.value`. `.set` will also return a `.noop` so `return cx.set(something)` will set the `computed.value` to `something` & not the return value.

### Async `computed` `cb` Code Example

```ts
let $person_id = atom()
let $person_op_count = atom(0)
let $search = atom()
let person_and_search_messages = computed(
  async cx => {
    if (!$person_id()) return null
    cx.set({ loading: true })
    $person_op_count.set($person_op_count.get() + 1) // .get() is called so no circular dependency
    let person = await fetch(`https://www.example.com/people/${$person_id()}`).then(response => response.json())
    let search_messages = await cx(() => { // push cx on stack
      return $search()
        ? fetch(`https://www.example.com/people/${$person_id()}/messages/${}`).then(response => response.json())
        : null
    }) // pop cx off stack
    return [person, search_messages]
  }
)
```

blikblum:
Thanks for the explanation. Now i see the benefits

btakita:
I implemented async support for the inline dependency `computed` function. The test is at https://github.com/btakita/nanostores/blob/issues/188/computed/computedSignal.test.ts#L365-L550

Total size-limit increases are:

- Atom: 313 B => 345 B (32 B increase)
- All: 1026 B => 1265 B (239 B increase)

.gitignore
```diff
@@ -1,3 +1,5 @@
 node_modules/
 
 coverage/
+
+.idea/
```
ai, line 5: Small code style fix: IDE files should be in a developer’s local git ignore

https://docs.github.com/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

btakita: Done! Thank you for the tip

btakita:
The latest commit wraps the predefined dependencies `computed` in a `cx`. This means the predefined dependencies `computed` will now auto-listen to atom functions & supports async.

I believe this PR is now ready for a full review.

btakita:
I did a bunch of size-limit optimizations & renamed `cx` to `autolisten`. Been iterating on this for a while & I think it's reaching a stable point for review. The original `computed` use cases should remain unaffected, while there is now autolisten & async support. If it looks good, I can squash the commits & we can merge into the main branch.

ai:
The current size is OK. But since it is a big PR which could break a lot, I will need more time to review it.

btakita:
Cool, I will test it on one of my projects as well.

btakita:
I'm applying the changes to my project & found a use case to add `.on` & `.off` to the `Autolisten` function.

Here is an example:

```ts
computed(al=>{
  const onchange = evt=>
    al.save(evt.matches ? 'dark' : 'light')
  let watch = window.matchMedia('(prefers-color-scheme: dark)')
  al.on(()=>
    watch.addEventListener('change', onchange)
  ).off(()=>
    watch.removeEventListener('change', onchange))
  return onchange(watch)
})
```

`.on` & `.off` are useful in that they automatically cleanup event listeners when the computed is off & when the `cb` is re-run due to a dependency change. It adds 44 B to the final size.

btakita:
I renamed autolisten to autosubscribe to match the semantics of subscribe which listens + immediately runs & returns the store value. `al` has been renamed to `as`.

btakita:
Another use case: checking existing value to guard against calling unnecessary async calls such as fetch using `autosubscribe()`

```ts
let $search = atom('')
let $searchMessages = computed(async asub => {
  if (!$search()) return null
  let response = fetch(`https://search.io/search/${encodeURIComponent($search())}`)
  let payload = await response.json()
  if (deepEqual(payload, asub())) return asub
  return payload
})
function deepEqual(val0, val1) {
  return JSON.stringify(val0) === JSON.stringify(val1)
}
```

btakita:
I made some changes. A new Autosubscribe is instantiated before a computed callback is run.

I added `.stale()` onto Autosubscribe to preserve reentrancy when an async callback is still running after subsequent callback begins.

A stale Autosubscribe has it's other methods "neutered". So .set, .save, .undo will not set the computed's value or `undoValue`, .on & .off will not run or save the callback function if the Autosubscribe is stale.

btakita:
I finished converting a medium-sized project (~several thousand stores) over to the autosubscribe computeds, which includes various async computeds & event binding within computeds. I also added support on a couple of libraries that extend `atom` & `computed`. Everything looks good on my end. There were some additions during the conversion process which includes:

- autosubscribe() (no arguments) returns the current value of the computed
- adding autosubscribe.stale() & .stale() checks to ensure reentrency support

I also added jsdocs for added use cases of `computed`.

I'm going to move onto another project, so let me know what else needs to be done.

atom/index.d.ts
```diff
@@ -99,6 +99,64 @@ export interface WritableAtom<Value = any> extends ReadableAtom<Value> {
 
 export type Atom<Value = any> = ReadableAtom<Value> | WritableAtom<Value>
 
+export interface Autosubscribe<Value = any> {
+  /**
+   * @return The computed store's value
+   */
+  (): Value
```
ai, line 106: I don’t see those change in `atom/index.js` and tests use `use()` method instead

atom/index.d.ts
```diff
@@ -99,6 +99,64 @@ export interface WritableAtom<Value = any> extends ReadableAtom<Value> {
...
+  /**
+   * Runs fn when the computed store is unmounted or when the computed store's cb is next run
+   *
+   * When .stale() is true, saving fn callback is deactivated.
+   *
+   * @param fn
+   */
+  off(fn: () => any): Autosubscribe<Value>
+  /**
+   * Immediately runs fn & calls fn when the computed store is mounted.
```
ai, line 122: We already have event system with `onMount`, `onStart` and `onStop`.

Do you avoid `onMount` for async support?

btakita: I did not know about `onStart` & `onStop`.

The issue with `onMount` is that it's passed the computed atom as an argument, which is not reliably accessible within the computed callback.

The use case is to be able to use `onMount`, `onStart`, & `onStop` from within the computed callback, so the computed's value can be directly set within the `onMount`. I added `on` & `off` not knowing about `onStart` & `onStop`. Perhaps `on` can be renamed to `onStart` & `off` can be renamed to `onStop`.

atom/index.d.ts
btakita: There are some conversations either incomplete or not yet had with the details of this PR. The PR has evolved from a the scope of just autosubscribing to including async & supporting some important edge cases. I hope to bring more clarity to these discussions.

> What is the different with .get()?

`.get` does not autosubscribe & keeps the same behavior that is currently in nanostores. 

> I don’t see those change in atom/index.js and tests use use() method instead.

> Can we use it only for autosubscriber()’s results?

`autosubscriber` should be renamed `autosubscribe`, as it is an instance of `Autosubscribe`.

Some context on the choise to use `use`...I'm looking for a short name to reduce api noise introduced by autosubscribe within the computed callback. Calling `autosubscribe($my-store)` is verbose, particularly when used in close proximity to other calls to `autosubscribe`. So I previously shortened `autosubscribe` to `asub`.

I started using `use()` in the tests in response to https://github.com/nanostores/nanostores/issues/188#issuecomment-1611611341. I'm hoping for us to agree on a convention. Note that `use` is also an object with member functions, such as `()`, `($atom: ReadableAtom)`, `.stale()`, `.set(val:Value)`, `.save(val:Value)`, `.on()`, `.off()`, etc.

The most terse api with the least noise is to make `atom` a function which autosubscribes the `atom` to the calling `computed`. I removed that behavior from the [PR branch](https://github.com/btakita/nanostores/tree/issues/188) in response to https://github.com/nanostores/nanostores/issues/188#issuecomment-1611701108. See [my reply](https://github.com/nanostores/nanostores/issues/188#issuecomment-1612032245) for more context.

The branch where atom is a function can be found at https://github.com/btakita/nanostores/tree/issues/188-atom-fn. I explained the change in [this comment](https://github.com/nanostores/nanostores/issues/188#issuecomment-1612032245). I can have this PR use the [issues/188-atom-fn](https://github.com/btakita/nanostores/tree/issues/188-atom-fn) branch if you agree with me on this point. If it were up to me, I would have this PR use the [issues/188-atom-fn](https://github.com/btakita/nanostores/tree/issues/188-atom-fn) branch, however I also do not want this to cause confusion by having both `$computed()` & `use($computed)` effectively be aliases to each other. I personally do not think it's confusing, but others may think it's confusing.

btakita: > I don’t see those change in atom/index.js and tests use use() method instead.
> Can we use it only for autosubscriber()’s results?

I also want to clarify the use case for the `Autosubscribe` `()` function. Inside the `computed` callback, there is no reliable way to have a reference to `$computed`. `()`, along with `.on()` & `.off()` (see the review topic below), support the use case of behaviorally accessing `$computed` within the computed callback.

btakita: > I personally think the terser $computed() is beneficial to reduce api noise...so I will use the atom function defined either in nanostores or in another library that extends nanostores

If we choose to not include the `$atom()` function in nanostores, perhaps a library which imports nanostores named "autostores" or "auto-nanostores" could be created to add the `$atom()` function behavior. This would allow the programmer to explicitly opt-into this behavior, bringing more awareness to how the implicit `autosubscribe` to `$atom()` behaves with it's benefits & inconsistencies.

atom/index.d.ts
btakita: On `Autosubscribe`, I renamed the methods `.on` to `.onStart` & `.off` to `.onStop`, keeping the names already used in the event system.

atom/index.d.ts
btakita: I renamed `autosubscriber` to `autosubscribe`.

atom/index.d.ts
ai: The problem is not in the name, but in duplication. Now we have the methods to do the same.

atom/index.d.ts
ai: The fact that using `.get()` will not make auto-subscriptions look strange for me. Some developer could forget it and use `get()` by a mistake and got an error which is very hard to debug.

atom/index.d.ts
btakita: Right. There are a few problems I ran into.

The existing `onStart` & `onStop` cannot be used inside of the computed callback:

- The `$computed` object is not accessible inside of the callback, which `onStart` & `onStop` requires
- The existing `onStart` & `onStop` have no stale checks so new `onStart` & `onStop` listeners would be bound every time the callback runs

This means the usage requires the `$computed` to be created first. Since `$computed` is a `ReadableAtom`, there is no way in the type system to directly set `$computed.value` or call `$computed.set`...In js, these props are there but a computed is not meant to be writable.

This means an `$atom` must be created which the `$computed` listens to.

Here is a comparison of the two approaches:

Using the existing `onStart` & `onStop`.

```ts
import { atom, computed, onStart, onStop } from 'nanostores'
let $colorSchemeAtom = atom()
let watch = window.matchMedia('(prefers-color-scheme: dark)')
let onchange = evt => $colorSchemeAtom(evt.matches ? 'dark' : 'light')
let $colorScheme = computed(use => use($atom))
onchange(watch)
onStart($computed, () => watch.addEventListener('change', onchange))
onStop($computed, () => watch.removeEventListener('change', onchange))
```

Using the autosubscribe.onStart & autosubscribe.onStop

```ts
import { computed } from 'nanostores'
let $colorScheme = computed(use => {
 let watch = window.matchMedia('(prefers-color-scheme: dark)')
 let onchange = evt =>use.save(evt.matches ? 'dark' : 'light')
 use.onStart(() => watch.addEventListener('change', onchange))
 use.onStop(() => watch.removeEventListener('change', onchange))
 return onchange(watch)
})
```

Having `.onStart` & `.onStop` in autosubscribe allows the `computed` callback to be completely self-contained without requiring the extra `$colorSchemeAtom` to be created & written to.

atom/index.d.ts
btakita: > Some developer could forget it and use get() by a mistake and got an error which is very hard to debug.

That is definitely something to consider with the ergonomics of using autosubsciptions.

I found that having a way to get the current state of a store without autosubscribing is useful in some cases, which will be more common with async computeds. This could be to avoid circular subscriptions causing a loop or referencing a store where a subscription is not wanted, such as a logger:

```ts
let $logger = atom()
let $userId = atom()
let $user = computed(use => {
  let userId = use($userId)
  if (userId == null) {
    info('No userId')
    return null
  }
  info('fetching userId: ', userId)
  return fetch(`https://my-api/users/${userId}`).then(response => response.json())
  function info(...argv) {
    let logger = $logger.get()
    if (logger) logger.info(...argv)
  }
})
```

Having the `$atom()` function would be helpful for these issues. Imagine `$atom()` being the primary way to get the value of the atom & autosubscribe, reserving `.get()` for special cases inside of a `computed` callback where an autosubscription is not wanted. That way, the normal (> ~98%) case is to simply call `$atom()` leaving `.get()` to be used in a small number of cases with the specific intent of getting the value without an autosubscription. In some of my codebases, I converted to using `$atom()` & noticed that `$atom()` is called > 98% of the time, with a few calls to `.get()` inside of some computed calbacks.

This leads to the other inconsistencies with using the `$atom()` function. The inconsistencies would only need to occur in an async `computed` callback when the synchronous task completes, which will be less often.

These inconsistencies could be documented with `computed`. So the idiom would be to use `$atom()` most of the time, `.get()` within a computed where autosubscribe is not wanted, & `use($atom)` or `$atom(use)` when the we want to autosubscribe after the synchronous task completes.

btakita: With that being said, I converted this pr branch `issues/188` to have the `$atom()` function. I created another branch where atom is not a function (the previous code for this PR) at https://github.com/btakita/nanostores/tree/issues/188-no-atom-fn.

atom/index.d.ts
btakita: An interesting impact of `.onStart`/`.onStop` inside of the computed callback is it enables a `$computed` to be driven without any dependencies on an `$atom`, bridging to non-nanostores event systems.

atom/index.d.ts
btakita: Another reason to keep the existing behavior of `.get` is preventing errors when upgrading to the version that contains this PR.

dtinth: Personally I agree about `.get()` not automatically subscribing. This makes it consistent with UI library bindings, where calling `.get()` inside a React component also won't make the component subscribe to the store, to subscribe `useStore` must be used instead of calling `.get()` directly.

atom/index.d.ts
ai: I also don’t like that current PR is huge and contain many different API changes :(. It is impossible to review and accept it in the current huge form.

btakita: > I don’t agree that it is more convenient.

I'm coming from an object pov with regards to less lines of code and self-containment vs multiple variables in the calling scope. The two example effectively do exactly the same thing from a declarative pov. There is a difference that I'll explain further down.

> computed() stores were created as derivatives stores from other stores. You are using them for any “smart” stores which calc values according to system API. I think it is better to separated derivatives, simple stores, and smart stores.

I don't see a need to create a distinction between derivative & smart stores. For example, what if you want both derivative & smart store behavior? Which one would be used? To me, I'm taking the term "computed" at face value. It is a computationally created atom. The computation can include autosubscribing derivations from other stores and/or listen to system events. Sure, there is historical usage, coming from Svelte I believe, of the derived store being derived from other stores. I think you made a brilliant choice to use "computed" instead. This PR just extends `computed` to a fuller meaning of the term "computed".

> computed() callback is calling on every parent-store changes.

Yes that is true. In this PR, if there are no parent stores, then the callback is called once.

> In your API store#onStart works like a React’s hook (works only first time, then ignored).

I'm not able to reconcile your interpretation with mine. We can further discuss to align our perspectives. The API is "re-entrant", meaning there is only binding of `.onStart`/`.onStop` at a time. Let's say the `computed` is both derived from another `$store` & uses `.onStart`/`.onStop`. When the parent `$store` updates, the callback is run again, which means `.onStart`/`.onStop` is called again. This api is re-entrant, so the previous binding of `.onStart`/`.onStop` is removed & then readded when the callback is invoked.

> It is not very convenient API.

It should effectively act the same as the existing API, except it allows `computed` to be self-contained without needing extra imperative logic & objects such as the `$colorSchemeAtom` along with the variables in the calling scope. Instead, the logic is self-contained within the `computed` callback, meaning the calling scope is not polluted with these imperative variables that only apply to the `$colorScheme`.

> I also don’t like that current PR is huge and contain many different API changes :(. It is impossible to review and accept it in the current huge form.

I hope you can reconsider. This PR does not have any breaking changes & is additive so existing usages of nanostores should not have any issues. I admit that this is a large change, but it's mainly to flesh out unsupported use cases such as autosubscription & async computeds in a cohesive way.

I bundled the `async` behavior along with the autosubscribe behavior to inform the design of the system. I & several others have bespoke functions to add async behavior to `computed`, so I think it is a relevant use case. If async `computeds` were available in nanostores, it will be more widely used because it's a powerful addition. Adding `async` behavior introduces the notions of staleness & re-entrancy, so this PR works through those issues.

Some convenience methods were added to `autosubscribe`:

- `()`: the current value of the `$computed` for deep-equal & diff checking
- `.set`/`.save` to support intermediate async states such as loading
- `.stale` to prevent old callback invocations to clobber the value of newer callback invocations
- `.onStart`/`.onStop` to support self-contained computeds using system apis

Having used this in my code-bases, I found all of these to be useful. `.onStart`/`.onStop` is not strictly necessary, but it does cut down on lines of code, & is self-contained meaning the functional pieces are all bundled within the callback. It also faithfully implements the term `computed` (that was a brilliant rename on your part) as a store that is computationally defined.

atom/index.d.ts
btakita: > This makes it consistent with UI library bindings, where calling .get() inside a React component also won't make the component subscribe to the store, to subscribe useStore must be used instead of calling .get() directly

Actually, autosubscribe would not occur if you use `$computed()` or `$computed.get`. Autosubscribe only occurs within the the synchronous task of a `computed` callback. However, I still think `.get` should keep it's existing behavior & not autosubscribe, so existing apps don't break.

btakita:
I took a long drive & was able to reflect on this PR. I acknowledge that it is very large, as @ai mentioned, & needs to be broken up into several more focused PRs with focused discussions. I also realized that async computeds is a breaking change. The returned `Promise`'s `.then` is called, which would break any implementations that rely on the `computed` returning a `Promise` object. So I'm going to close this & create the following PRs.

- Package size optimizations
  - in an effort to keep the size impact of this PR as small as possible, there are some size optimizations
- Solidjs-like synchronous `computed`
  - This will include the atom function & autosubscription. No `autosubscribe` function object
- async `computed`
  - This would be a breaking change as the `Promise` `.then()` would be run. It's possible to come up with an alternative set of implementation or identify patterns which allow a `Promise` to be returned.
  - This includes an implementation of the `autosubscribe` function object
    - I would like to rename `autosubscribe` to `task` or `computedTask`, as it does more than support autosubscribe.
    - Included non-controversial methods would include:
      - `.()`: returns the current value of the `$computed`
      - `.($store)`: autosubscribes the `$computed` to the `$store`
      - `.set(val)`: sets `val` as an intermediate (undoable) `value`
      - `.undo()`: undo the intermediate `value` to the last saved value
      - `.save(val)`: sets `val` to the `value` & `undoValue`
        - A call to `.undo()` will reset to `value` to `val`. Note that the return value of the `computed` callback, which is not a `Task`, is saved. If the `computed` return's it's task, the return value is ignored
      - `.stale()`: return true if another `task` is started after the `task` in the run of the `computed` callback
        -  If `.stale()` is true, then calls to `.set`, `.save`, & `.undo` do not change the `value`. This ensures that only the current `task` is able to set `value`.
        - `.stale()` also allows the opportunity to not make expensive system api calls
- Controversial additions with more focused discussions about their merits & demerits
  - `.onStart(cb)`/`.onStop(cb)`
    - While these require less LoC & are self-contained compared to the existing `onStart`/`onStop`, the runtime behavior is different & may not be worth adding.
  -  Private stores
    - I added these before `.stale()` was added as a possible way to handle certain use cases such as cancelling an async computed task. With `.stale()` added, it is not necessary. There may be some possible use cases, but overall I think there may be better idioms to solve some, if not all, of these use cases. I'll add more detail in the PR.

PR #204, opened 2023-06-06
removing notifyId comparison from computed storage

aleksandrjet:
I don't want to do difficult calculations with the same parameters. I think, that changing parameters is the only reason to recalculate computed store


ai:
We already removed it

PR #214, opened 2023-07-03
types(map): extend `MapCreator`

euaaaio:
These changes are needed to make the `build` method a standard for creating store builders.

### DIFF
--- a/map/index.d.ts
+++ b/map/index.d.ts
@@ -131,6 +131,7 @@ export interface MapCreator<
   Args extends any[] = []
 > {
   (id: string, ...args: Args): MapStore<Value>
+  build(id: string, ...args: Args): MapStore<Value>
   cache: {
     [id: string]: MapStore<Value & { id: string }>
   }


euaaaio:
So we can make a universal logger for store builders: https://github.com/nanostores/logger/blob/592becd0de83334d212e73af4b97fd9cb69a8b37/creator-logger/index.js

ai:
It will be the major release (maybe we even need a helper back).

I will review https://github.com/nanostores/nanostores/pull/203 first

euaaaio:
It’s ok. For now, in Logger we check the stores for the `cache` and `build` properties to detect if it is a `MapCreator`.

ai:
I took back `mapCreator()` but this type with new name and with semi-private style (we will not promote it in docs) https://github.com/nanostores/nanostores/commit/8086e3795780b9151f1c4814860a6f0441b2a8a7

ai:
I am planning to release it together with Context.

I am planing to move `Creator.cache` to context.

PR #221, opened 2023-07-16
implement synchronous autosubscribe (#188)

btakita:
+ BoxTask,task,Task,UnboxTask

atom: is a function which takes an optional autosubscribe function & returns .get()

computed:

	+ task: (): returns computed store's value (store: AnyStore): computed store autosubscribes to the given store

package.json: description: + SolidJS

Thanks to @dtinth for autosubscribe function type inference which he implemented in https://github.com/dtinth/nanostores-computed-dynamic

size-limit:

	Atom: + 26 B
	All: + 143 B


btakita:
Ported from https://github.com/nanostores/nanostores/pull/203

btakita:
In this PR;

- `Autosubscribe` renamed to `Task`
- `autosubscribe` renamed to `getTask`

btakita:
Updated the idiomatic usage of `$store()` over `$store.get()` in the docs.

Also force pushed a fix to `action` including the `$store` function object.

btakita:
Force pushed an addition to the `Task` interface:

```ts
export interface Task<Value = any> {
  // ...
  <V>($store: AnyStore<V>): V
  // ...
}
```

ai:
Where we are planning to use `Task` in additional to `autosubscribe()` computed store?

btakita:
> Where we are planning to use `Task` in additional to `autosubscribe()` computed store?

I'm proposing using `Task` instead of `Autosubscribe` & `getTask` instead of `autosubscribe`.

ai:
I spoke with a few users of Nano Stores and heard the same concern, which I have. I worry that very simple and explicit API of Nano Stores is becoming tricky and too magical.

The internal complexity of the system grows fast, but we have only small practical reasons from this API (I agree that autosubscribing is cool, but it doesn’t give you enough guarantee and performance benefits are small).

I know that you already did 4 (!) PR, and it is huge of work. Thanks for that. Without a prototype it was hard to understand how it could affect the project. Excuse us if we will close the PR.

This is only my current thoughts. I am open for opinion from another persons.

dtinth:
I would _really_ love to see support for **computed stores with dynamic dependencies** built in to Nano Stores (so that I don’t have to maintain my userland solution [nanostores-computed-dynamic](https://github.com/dtinth/nanostores-computed-dynamic)), and this PR is a step towards that direction.

However, I agree that, in its current form, it makes the API more complex. I am not a fan of the additional syntax for reading store (`$store()`) or autosubscribing to a store (`$store(task)`) and prefer that there is only one way to read from a store (`$store.get()`) or subscribing (`use($store)`).

btakita:
> I spoke with a few users of Nano Stores and heard the same concern, which I have. I worry that very simple and explicit API of Nano Stores is becoming tricky and too magical.
> The internal complexity of the system grows fast, but we have only small practical reasons from this API (I agree that autosubscribing is cool, but it doesn’t give you enough guarantee and performance benefits are small).
> I know that you already did 4 (!) PR, and it is huge of work. Thanks for that. Without a prototype it was hard to understand how it could affect the project. Excuse us if we will close the PR.

I missed this comment. If you want to go ahead without this PR, I will fork this project. Probably call it autostores. I'll try to tread nanostores as an upstream project if the effort is small. That way, if you decide to go with it, I can deprecate autostores.

ai:
> Probably call it autostores. 

I will promote it in my social media 👍

btakita:
> I will promote it in my social media +1

I appreciate it! I'm going through the exercise of creating the fork & realized the breadth of the ecosystem. I may end up creating some extension functions that support autosubscribe & convert existing `$stores` to autosubscribe. I also noticed nanostores already has a `task` function :pensive:. I'll need some time to familiarize myself more with the full api.

@ai What are your thoughts on @dtinth's approach?

btakita:
I'm going to try to strip down the PR until it is below the complexity threshold where everyone is comfortable. In the latest force push, I made the following changes.

The private `taskStack` is renamed to `computedTaskStack` to avoid the name collision with the existing `task` export.

`getTask` was removed.

The `ReadableAtom` type `.()` method no longer takes a `computedTask` as an argument. So now there is only one way to autosubscribe, which is to use `$store()`.

Removing the `computedTask` from user-land removes the need for the `BoxTask` & `UnboxTask` types, so those were removed. Also, the `computed` callback can no longer return the `computedTask` so that check was removed.

@ai @dtinth How does the complexity of the PR look now? Is the `$store()` idiom palatable?

btakita:
I'm going to try to strip down the PR until it is below the complexity threshold where everyone is comfortable. In the latest force push, I made the following changes.

The private taskStack is renamed to autosubscribeStack to avoid the name collision with the existing task export.

getTask was removed.

The ReadableAtom type .() method no longer takes a task/autosubscribe as an argument. So now there is only one way to autosubscribe, which is to use $store().

Removing the task/autosubscribe from user-land removes the need for the BoxTask & UnboxTask types, so those were removed. Also, the computed callback can no longer return the task/autosubscribe so that check was removed.

The internal computed autosubscribe function only takes a $store as an argument, simplifying it's implementation.

@ai @dtinth How does the complexity of the PR look now? Is the $store() idiom palatable?

dtinth:
I have one concern about the `$store()` syntax: For it to work, it depended on an internal, module-level `autosubscribeStack` variable.

This is fine, but if some 3rd-party library ships with a different version of Nano Stores as a dependency, then are is a chance that multiple copies of `nanostores` would exist on the same page. Each version has a copy of `autosubscribeStack`.

In this case, the atom from one version wouldn’t be auto-subscribable from another version’s `computed()` function, i.e., they are not inter-operable.

For example:

```js
// Note: This package doesn’t exist - it is just an example,
// but it reflects how I use Nano Stores in my app.
//
// Let’s assume that this package ships with its own version of
// Nano Stores, either via an incompatible dependency version,
// or *gasp* bundling it into `dist/index.js`
//
import { $currentUser } from '@example/nanostores-firebase'

// Another copy of of `nanostores` is imported.
//
import { computed } from 'nanostores'

// Autosubscription would fail because `$currentUser` is
// associated with a different copy of `autosubscribeStack`.
//
const $displayName = computed(() => $currentUser()?.displayName)
```

Having multiple copies of the same library can happen when importing modules in bundle-less scenarios such as `esm.run`.

btakita:
> This is fine, but if some 3rd-party library ships with a different version of Nano Stores as a dependency, then are is a chance that multiple copies of nanostores would exist on the same page. Each version has a copy of autosubscribeStack.

Excellent observation! I could create a global variable using `globalThis`, but it will add ~31 B to the size. We may be able to use the [upcoming context feature](https://github.com/nanostores/nanostores/pull/228) to attach to.

btakita:
I ended up implementing [rmemo](https://github.com/ctx-core/rmemo) which has autosubscribe behavior built in.

PR #222, opened 2023-07-16
Async computed

btakita:
computed: async support (#188)

  - BREAKING CHANGE: If the `$computed()` returns a `PromiseLike`, the `PromiseLike`'s `.then()` is called to set `$computed.value`.
  - Mitigation: the `$computed()` can return an object wrapping the `PromiseLike`.
    - During the synchronous run of the `computed` callback, the `task` is pushed onto the `taskStack`.

Task:

  + save(newValue: Value): Task<Value>
  + set(intermediateValue: Value): Task<Value>
  + stale(): boolean
  + undo(): Task<Value>

size-limit:

  - All: + 78 B


btakita:
Ported from https://github.com/nanostores/nanostores/pull/203

ai:
After the consideration I decided to close this PR as well.

It is comming with API changes and I afraid that it is overcompilcate the simple nature of Nano Stores.

It is still a great design for another project where this API will be initialy part of the design.

PR #224, opened 2023-07-17
Adding batch-mode for `computed`

dkzlv:


computed/index.d.ts
```diff
@@ -13,7 +13,8 @@ type C = (...values: StoreValues<[A, B]>) => void
 interface Computed {
   <Value extends any, OriginStore extends Store>(
     stores: OriginStore,
-    cb: (value: StoreValue<OriginStore>) => Value
+    cb: (value: StoreValue<OriginStore>) => Value,
+    batched?: boolean
```
ai, line 17: Should we use `{ batched: true }` to improve readability (without knowing this rare option it will be hard to understand what is `true`)?

In one hand it will increase bundle size. But if it is only a few bytes, we can go there.

README.md
```diff
@@ -296,11 +296,39 @@ You can combine a value from multiple stores:
...
+export const $link = computed(
+  [$sortBy, $categoryIdFilter],
+  (sortBy, categoryId) => {
+    return `/api/entities?sortBy=${sortBy}&categoryId=${categoryId}`
+  },
+  // Notice this argument!
+  true
+)
+
+/**
```
ai, line 323: Let’s use `//` comment here because `/**` is JSDoc/TSDoc.

computed/index.d.ts
dkzlv: Yeah, I was thinking about that too. 

My another idea was to make this arg a private one, but introduce a separate export `batched` that sets it privately. Thing is, the more I think about the more I'm convinced that I'll set it to `true` always, no matter what. Like, I don't see a case when I need it to be `false`. And in that case it's just a mental overload to remember to never miss it.

ai: I mean `computed(prev, cb, { batched: true })` instead of `computed(prev, cb, true)`

ai:
I merged it to `next` branch. I am planning to release it together with `Context`.

PR #226, opened 2023-07-19
Variables in tests & jsdocs match $ prefix convention

btakita:
ported from https://github.com/nanostores/nanostores/pull/221


ai:
Thanks!

PR #228, opened 2023-07-19
Add contexts

dkzlv:
Super mega WIP, but outlines the general look and feel

To do:
- [x] test all lifecycle hooks properly
- [x] test all places I touched and if they work properly with different contexts (`map`, `deepMap`)
- [x] things that won't work with different contexts 100%:
    - [x] `computed`
    - [x] `task`
    - [x] `action`
- [x] we need an understanding of how context will be serialized and how atom will be hydrated
- [ ] look for ways to optimize size penalty for this feature
- [ ] readme updates, jsdoc documentation, etc.

### DIFF
--- a/action/index.js
+++ b/action/index.js
@@ -12,7 +12,7 @@ export let actionId = Symbol()
 export let doAction = ($$store, actionName, cb, args, ctx = globalContext) => {
   let taskContext = ensureTaskContext(ctx)
 
-  let id = ++taskContext.i
+  let id = ++taskContext.id
   let $store = withContext($$store, ctx)
 
   let tracker = { ...$store }
--- a/action/index.js
+++ b/action/index.js
@@ -1,12 +1,19 @@
+import {
+  ensureTaskContext,
+  globalContext,
+  withContext
+} from '../context/index.js'
 import { startTask } from '../task/index.js'
 
 export let lastAction = Symbol()
 export let actionId = Symbol()
 
-let uid = 0
+export let doAction = ($$store, actionName, cb, args, ctx = globalContext) => {
+  let taskContext = ensureTaskContext(ctx)
 
-export let doAction = ($store, actionName, cb, args) => {
-  let id = ++uid
+  let id = ++taskContext.i
+  let $store = withContext($$store, ctx)
+
   let tracker = { ...$store }
   tracker.set = (...setArgs) => {
     $store[lastAction] = actionName
@@ -30,7 +37,7 @@ export let doAction = ($store, actionName, cb, args) => {
   }
   let result = cb(tracker, ...args)
   if (result instanceof Promise) {
-    let endTask = startTask()
+    let endTask = startTask(ctx)
     return result
       .catch(error => {
         if (onError) onError(error)
--- a/action/index.js
+++ b/action/index.js
@@ -1,6 +1,7 @@
 import {
   ensureTaskContext,
   globalContext,
+  isContext,
   withContext
 } from '../context/index.js'
 import { startTask } from '../task/index.js'
@@ -40,19 +41,24 @@ export let doAction = ($$store, actionName, cb, args, ctx = globalContext) => {
     let endTask = startTask(ctx)
     return result
       .catch(error => {
-        if (onError) onError(error)
+        onError?.(error)
         throw error
       })
       .finally(() => {
         endTask()
-        if (onEnd) onEnd()
+        onEnd?.()
       })
   }
-  if (onEnd) onEnd()
+  onEnd?.()
   return result
 }
 
 export let action =
   ($store, actionName, cb) =>
-  (...args) =>
-    doAction($store, actionName, cb, args)
+  (...args) => {
+    let argsToSpread = [args]
+    let possiblyCtx = args[args.length - 1]
+    if (isContext(possiblyCtx)) argsToSpread.push(possiblyCtx)
+
+    return doAction($store, actionName, cb, ...argsToSpread)
+  }
--- a/atom/index.d.ts
+++ b/atom/index.d.ts
@@ -95,14 +95,6 @@ export interface WritableAtom<Value = any> extends ReadableAtom<Value> {
    * @param newValue New store value.
    */
   set(newValue: Value): void
-
-  /**
-   * Low-level method to read store’s value without calling `onStart`.
-   *
-   * Try to use only {@link ReadableAtom#get}.
-   * Without subscribers, value can de undefined.
-   */
-  value: undefined | Value
 }
 
 export type Atom<Value = any> = ReadableAtom<Value> | WritableAtom<Value>
--- a/atom/index.d.ts
+++ b/atom/index.d.ts
@@ -95,6 +95,14 @@ export interface WritableAtom<Value = any> extends ReadableAtom<Value> {
    * @param newValue New store value.
    */
   set(newValue: Value): void
+
+  /**
+   * Low-level method to read store’s value without calling `onStart`.
+   *
+   * Try to use only {@link ReadableAtom#get}.
+   * Without subscribers, value can de undefined.
+   */
+  value: undefined | Value
 }
 
 export type Atom<Value = any> = ReadableAtom<Value> | WritableAtom<Value>
--- a/atom/index.d.ts
+++ b/atom/index.d.ts
@@ -52,6 +52,16 @@ export interface ReadableAtom<Value = any> {
   listen(listener: (value: ReadonlyIfObject<Value>) => void): () => void
 
   /**
+   * Name of this store.
+   *
+   * Its primary use is SSR: `name` should uniquely identify this atom. If
+   * name is not set, this atom won't participate in SSR state (de-)serialization.
+   *
+   * It can also be used in various developer tools (e.g., browser extensions).
+   */
+  name?: string
+
+  /**
    * Unbind all listeners.
    */
   off(): void
--- a/atom/index.js
+++ b/atom/index.js
@@ -3,7 +3,7 @@ import { getStoreState, globalContext } from '../context/index.js'
 
 let uid = 0
 
-export let atom = (initialValue, level) => {
+export let atom = (initialValue, level = 0) => {
   let $atom = {
     ctx: globalContext,
     get() {
@@ -15,7 +15,7 @@ export let atom = (initialValue, level) => {
     },
     id: uid++,
     iv: initialValue,
-    l: level || 0,
+    l: level,
     get lc() {
       let state = getStoreState(this, $atom)
       return state.lc
@@ -28,7 +28,7 @@ export let atom = (initialValue, level) => {
         let index = state.ls.indexOf(listener)
         if (~index) {
           state.ls.splice(index, 2)
-          if (!--state.lc) this.off()
+          if (!--state.lc) this.off?.()
         }
       }
     },
@@ -61,8 +61,6 @@ export let atom = (initialValue, level) => {
         listenerQueue.length = 0
       }
     },
-    off() {} /* It will be called on last listener unsubscribing.
-                We will redefine it in onMount and onStop. */,
     set(data) {
       let state = getStoreState(this, $atom)
       if (state.v !== data) {
@@ -90,7 +88,7 @@ export let atom = (initialValue, level) => {
       let state = getStoreState(this, $atom)
       state.ls = []
       state.lc = 0
-      $atom.off()
+      $atom.off?.()
     }
   }
 
--- a/atom/index.js
+++ b/atom/index.js
@@ -1,37 +1,38 @@
 import { clean } from '../clean-stores/index.js'
-import { getListenerQueue, getStoreState } from '../context/index.js'
+import { getStoreState, globalContext } from '../context/index.js'
 
 export let atom = (initialValue, level) => {
   let $atom = {
+    ctx: globalContext,
     get() {
-      let state = getStoreState($atom)
+      let state = getStoreState(this, $atom)
       if (!state.lc) {
-        $atom.listen(() => {})()
+        this.listen(() => {})()
       }
       return state.v
     },
     iv: initialValue,
     l: level || 0,
     get lc() {
-      let state = getStoreState($atom)
+      let state = getStoreState(this, $atom)
       return state.lc
     },
-    listen(listener, listenerLevel) {
-      let state = getStoreState($atom)
-      state.lc = state.ls.push(listener, listenerLevel || $atom.l) / 2
+    listen: function (listener, listenerLevel) {
+      let state = getStoreState(this, $atom)
+      state.lc = state.ls.push(listener, listenerLevel || this.l) / 2
 
       return () => {
         let index = state.ls.indexOf(listener)
         if (~index) {
           state.ls.splice(index, 2)
           state.lc--
-          if (!state.lc) $atom.off()
+          if (!state.lc) this.off()
         }
       }
     },
     notify(changedKey) {
-      let state = getStoreState($atom)
-      let listenerQueue = getListenerQueue()
+      let state = getStoreState(this, $atom)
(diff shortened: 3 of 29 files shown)


dkzlv:
Acknowledge: I basically backported the whole branch by @TrySound in #199.

atom/index.js
```diff
@@ -1,81 +1,93 @@
 import { clean } from '../clean-stores/index.js'
-
-let listenerQueue = []
+import { getStoreState, globalContext } from '../context/index.js'
 
 export let atom = (initialValue, level) => {
-  let listeners = []
   let $atom = {
+    ctx: globalContext,
```
ai, line 6: _Sorry, I will have dump question. It is a trade-off of quick review (like SAW vs DOM parsers)_

Why do we need to bind global context to every context?

lifecycle/index.js
```diff
@@ -122,34 +132,34 @@ export let onMount = ($store, initialize) => {
...
-    $store.off = () => {
-      originOff()
+    $store.off = function () {
+      originOff()?.call(this)
       setTimeout(() => {
-        if ($store.active && !$store.lc) {
-          $store.active = false
-          for (let destroy of $store.events[UNMOUNT]) destroy()
-          $store.events[UNMOUNT] = []
+        if (this.active && !this.lc) {
```
ai, line 148: Got about `this` trick. Yes, it is OK.

atom/index.js
```diff
@@ -1,81 +1,93 @@
...
       }
-      return $atom.value
+      return state.v
     },
+    iv: initialValue,
     l: level || 0,
-    lc: 0,
-    listen(listener, listenerLevel) {
-      $atom.lc = listeners.push(listener, listenerLevel || $atom.l) / 2
+    get lc() {
```
ai, line 16: Nice 😈

context/index.js
```diff
@@ -0,0 +1,89 @@
...
+export let resetContext = name => {
+  if (name) {
+    contexts.delete(name)
+  } else {
+    contexts.clear()
+    globalContext.d = false
+    delete globalContext.ta
+  }
+}
+export let getContext = name => contexts.get(name)
```
ai, line 31: Should we use `function` style everywhere (like we use below)?

context/index.js
```diff
@@ -0,0 +1,89 @@
...
+// lazily initialize store state in context
+export function getStoreState(thisStore, originalStore) {
+  if (thisStore.ctx.d) {
+    if (process.env.NODE_ENV !== 'production') {
+      throw new Error(
+        `You can't mix global context and custom contexts, that is probably an error. ` +
+          'Please, pass the correct context into `withContext`.'
+      )
+    }
+    // eslint-disable-next-line no-throw-literal
```
ai, line 42: We need a comment here about how we are planning to use it.

Also, just disable `no-throw-literal` globally if we are going to use this signal method.

context/index.js
```diff
@@ -0,0 +1,89 @@
...
+    contexts.clear()
+    globalContext.d = false
+    delete globalContext.ta
+  }
+}
+export let getContext = name => contexts.get(name)
+
+// lazily initialize store state in context
+export function getStoreState(thisStore, originalStore) {
+  if (thisStore.ctx.d) {
```
ai, line 35: `d` = default?

context/index.js
```diff
@@ -0,0 +1,89 @@
...
+  if (!state) {
+    state = {
+      // [l]isteners [c]ount
+      lc: 0,
+      // [l]isteners
+      ls: [],
+      // [v]alue
+      v: thisStore.iv
+    }
+    thisStore.ctx.s.set(originalStore, state)
```
ai, line 56: I'm afraid we are going to be cursed for that one-letter code. Maybe we are not soooo into optimization to going for that black magic?

What about rule use it only when “all our colleagues can get what this shortcut means”?

context/index.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+  return state
+}
+
+function shallowClone(obj) {
+  return Object.create(
```
ai, line 62: Nice! The deep JS knowledge.

context/index.js
```diff
@@ -0,0 +1,89 @@
+let contexts = new Map()
+
+let _createContext = name => {
```
ai, line 3: `_` is too encrypting for the place where minifier will change names.

What about `buildContext` and `createContext` (`build` is about building the variable and `create` is about side effects)?

context/index.d.ts
```diff
@@ -0,0 +1,17 @@
+import type { ReadableAtom } from '../atom'
+
+// This isn't a part of the public API for now
+type Context = unknown
```
ai, line 4: We need a public API for dev tools:

1. Get all active stores on a current page
2. Including all stores from store creators like `User(id)` from Logux Client or [MapCreator](https://github.com/nanostores/nanostores/blob/next/map-creator/index.d.ts)

context/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+export type ContextGetter = <T extends ReadableAtom>(store: T) => T
+export type WithCtx = {
+  ctx: ContextGetter
+}
+
+export declare const globalContext: Context
+
+export function createContext(name: string): Context
+export function resetContext(name?: string): void
+export function getContext(name: string): Context | undefined
```
ai, line 15: Why do we need context name? Can we just use `id` from `createContext`?

context/index.js
ai: Whta about global marker like `let multipleContext = false`?

atom/index.js
```diff
@@ -1,81 +1,93 @@
...
 
       return () => {
-        let index = listeners.indexOf(listener)
+        let index = state.ls.indexOf(listener)
         if (~index) {
-          listeners.splice(index, 2)
-          if (!--$atom.lc) $atom.off()
+          state.ls.splice(index, 2)
+          state.lc--
+          if (!state.lc) this.off()
```
btakita, line 29: This will shave off ~3 B.

```js
if (!--state.lc) this.off()
```

atom/index.js
dkzlv: I suppose we could, but it'll be in more places. Currently `this.ctx` appears 4 times, so this is just 4 times repeated the same `.ctx || globalContext`. Seems easier to me to just add context to all stores by default. Do you see any downsides?

atom/index.js
dkzlv: That's Bogdan's idea! Yeah, it's good.
cc @TrySound

context/index.js
dkzlv: Tbh, I suspect it shouldn't fire at all. We have an elaborate error message for development, but it seemed to me that it shouldn't change the behavior in production. But I wanted the throw to affect bundle size as little as possible. I think, throwing a number is the smallest in terms of chars. I don't expect anyone to catch it or work with it in any way.

context/index.js
dkzlv: I was just following the convention 😂 Yeah, I agree, it's incomprehensible, I'll change that.

context/index.d.ts
dkzlv: Yeah, sure. That's a TBD.
The way I currently store the assoc between store and its value is `Map<Store, StoreValue>`. Main problem is that it's not _serializable_. I don't have a solution for this yet, but that will definitely change the way we store contexts, hence change the public API for contexts and context getters.

context/index.d.ts
dkzlv: Yep, that's the same id :) Well, it's called `name` for now, but I'm open to changing it to `id`, that seems even better.

context/index.js
ai: In this case I think it is better to throw `new Error('no global context')` or `new Error('CODE')`

context/index.d.ts
```diff
@@ -0,0 +1,18 @@
...
+type Context = unknown
+
+export type ContextGetter = <T extends ReadableAtom>(store: T) => T
+export type WithCtx = {
+  ctx: ContextGetter
+}
+
+export declare const globalContext: Context
+
+export function createContext(id: string, storeValues?: object): Context
```
ai, line 13: А зачем нам `id: string`? Так надо думать над `id`

dkzlv: That's a good question. I don't know!

I suppose, there are 2 main reasons I see immediately:
1. if there's no persistent `id`, it means that each time you call `createContext` a new context is created. This doesn't really conform with the reality of frameworks (which typically rerender like crazy)
2. we'll eventually need to serialize the context into a string. If the dev doesn't know the ID, they won't be able to serialize the right context.

Nonetheless, I suppose we'll find out whether we actually need this arg once we try to use them with an SSR-backed framework for the first time. Don't have any concrete answers right now.

ai: Let’s remove `id: string` for now to save the space

dkzlv: But rerenders? 😄 

```tsx
const App = ({children}) => {
  const [state, set] = useState(0)

  return <>
    <button onClick={() => set(state + 1)}>inc</button>
    <NanostoresContext value={createContext()}>{children}</NanostoresContext>
  </>
}
```

If you don't have any idea on how to retain its identity between rerenders (remember, `useMemo` won't work), it's better leave it as is until we have a proof it won't be needed. In the end we won't publish this code without first writing framework (and even meta-framework) specific code that consumes it. This whole PR is just an idea of what should be enough to make it work, but the reality can be tough.
If I find that it's not really required, I'll drop it then, but right now I have a strong feeling it's required.

ai: Why not

```js
const ctx = createContext()

const App = ({children}) => {
  const [state, set] = useState(0)

  return <>
    <button onClick={() => set(state + 1)}>inc</button>
    <NanostoresContext value={ctx}>{children}</NanostoresContext>
  </>
}
```

As I understand this file will be executed for every user and each will have unique context.

ai: I mean:

```jsx
const ctx = createContext()

const App = ({children}) => {
  return <>
    <NanostoresContext value={ctx}>{children}</NanostoresContext>
  </>
}
render(<App />, domNode)
```

We can wrap it into the function to be sure that context will be created for every user

action/index.js
```diff
@@ -1,12 +1,20 @@
...
+  isContext,
+  withContext
+} from '../context/index.js'
 import { startTask } from '../task/index.js'
 
 export let lastAction = Symbol()
 export let actionId = Symbol()
 
-let uid = 0
+export let doAction = ($$store, actionName, cb, args, ctx = globalContext) => {
```
ai, line 12: Maybe context API fort actions will look like:

```js
renamePost(withContext($post), 'New title')
```

ai: I prefer to have single API for contexts and using latest argument can be tricky

atom/index.d.ts
```diff
@@ -95,6 +95,14 @@ export interface WritableAtom<Value = any> extends ReadableAtom<Value> {
...
    */
   set(newValue: Value): void
+
+  /**
+   * Low-level method to read store’s value without calling `onStart`.
+   *
+   * Try to use only {@link ReadableAtom#get}.
+   * Without subscribers, value can de undefined.
+   */
+  value: undefined | Value
```
ai, line 105: We already have it in `ReadableAtom`. Why do we need it here too?

atom/index.js
```diff
@@ -1,82 +1,94 @@
 import { clean } from '../clean-stores/index.js'
+import { getStoreState, globalContext } from '../context/index.js'
 
-let listenerQueue = []
+let uid = 0
```
ai, line 4: `uid` is used for `unique id`.

Let’s use `lastId` to explain the it is a counter.

action/index.js
dkzlv: @ai Yeah, that can be done, sure. Only thing is that won't change the broken contract with the last argument for `action`. I couldn't come up with a better approach, tbh. Any ideas?

atom/index.d.ts
dkzlv: That's strange, I have no idea how it got here. Will ✂️ it

context/index.d.ts
ai: The new snippet:

```js
function renderForClient () {
  let ctx = createContext()
  const App = ({children}) => {
    return <>
      <NanostoresContext value={ctx}>{children}</NanostoresContext>
    </>
  }
  render(<App />, domNode)
}
```

context/index.js
```diff
@@ -0,0 +1,115 @@
...
+}
+
+let globalContextPolluted = false
+export const globalContext = buildContext()
+
+export function createContext(storeStates) {
+  globalContextPolluted = true
+  return buildContext(storeStates)
+}
+export function resetContext(context) {
```
ai, line 31:
```suggestion

export function resetContext(context) {
```

context/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export type ContextGetter = <T extends ReadableAtom>(store: T) => T
+export type WithCtx = {
+  ctx: ContextGetter
+}
+
+export declare const globalContext: Context
+
+export function createContext(storeValues?: object): Context
+export function resetContext(context?: Context): void
```
ai, line 14: Can’t understand by function name what does it do.

context/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+type Context = unknown
+
+export type ContextGetter = <T extends ReadableAtom>(store: T) => T
+export type WithCtx = {
+  ctx: ContextGetter
+}
+
+export declare const globalContext: Context
+
+export function createContext(storeValues?: object): Context
```
ai, line 13: We will need better types for that to be able to use it in mocks.

I think we need to plan context API for mocks and tests on the current state of PR.

context/index.d.ts
```diff
@@ -0,0 +1,17 @@
+import type { ReadableAtom } from '../atom'
+
+// This isn't a part of the public API for now
+type Context = unknown
+
+export type ContextGetter = <T extends ReadableAtom>(store: T) => T
+export type WithCtx = {
```
ai, line 7: Too simple type to be exported. Let’s just use `& { ctx: ContextGetter }` in lifecycle

context/index.d.ts
```diff
@@ -0,0 +1,17 @@
+import type { ReadableAtom } from '../atom'
+
+// This isn't a part of the public API for now
+type Context = unknown
+
+export type ContextGetter = <T extends ReadableAtom>(store: T) => T
```
ai, line 6: Can we rename type to something like `ContextInjector` to make it clear what it does

context/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  ctx: ContextGetter
+}
+
+export declare const globalContext: Context
+
+export function createContext(storeValues?: object): Context
+export function resetContext(context?: Context): void
+export function serializeContext(context: Context): string
+
+export function withContext<T>(storeOrAction: T, ctx: Context): T
```
ai, line 17: I think we can also start writing docs in TSDoc. The public API looks good, but docs will help us to see the bigger picture.

bernatfortet:
@ai did you stop working on this? or am I missing something?

ai:
cc @dkzlv

ai:
@dkzlv stopped to work on this feature.

We needed API to set store and a few small changes.

I am not sure why we closed PR, let’s re-open it.

sia-fl:
Has this feature been discontinued? Is anyone still working on this feature?

lucasmotta:
Hey @ai I would love to help get this out there. Maybe with the docs or code review?

ai:
@lucasmotta sure! The plan:

1. Create new PR from this one.
2. Check all my comments in this PR and fix them.
3. Add API to mock stores for tests.
4. Rebase to the new `next` branch (better to do it on Saturday, I want to merge one PR).
5. Add docs

PR #229, opened 2023-07-20
Async computed using `task` without autosubscribe

btakita:
This version of the `computed` allows `task` to be used to enable async operations.

```js
import { computed, task } from 'nanostores'

import { $userId } from './users.js'

export const $user = computed($userId, userId => task(async () => {
  const response = await fetch(`https://my-api/users/${userId}`)
  return response.json()
}))
```


btakita:
See https://github.com/orgs/nanostores/discussions/231

btakita:
[Related discussion](https://github.com/orgs/nanostores/discussions/216)

ai:
Looks good. I will try to accept it after https://github.com/nanostores/nanostores/pull/228

ai (approved):

ai:
Thanks a lot! Great and simple addition.

ai:
I just found that async computed types is broken. Fixed with afa972a

PR #232, opened 2023-07-25
Local contexts

dkzlv:
To do:

- [ ] serialize/deserialize nested contexts + tests

The basic problem this PR aims to solve is that we don't have a feasible way to work with local component state. Say, you want to have a form component that can have multiple independent instances at a time. Generic contexts could help you with this, but they create a completely _new_ space, therefore limiting your ability to use atoms from the higher contexts.

This is where Local Contexts come into play. This is a hidden core feature which most users don't need to know about. Instead they will interact with it in the framework space. So, pseudo-code time!

```tsx
/**
 * 2 new exports in all framework-related libs:
 * 1. `NanostoresContext`. Component that handles context switches. More about it later
 * 2. `useCtx`. Returns the `ctx` wrapper that's used for cases like setting an input value
 *
 * For the purposes of gradual adoption both are actually optional: if you don't use
 * `NanostoresContext`, it'll take the default value (global context), same applies to
 * direct `$store.set()` calls.
 *
 * These two are not local context-specific, they are a requirement for contexts in general.
 */
import { useStore, useCtx, NanostoresContext } from "@nanostores/react";
import { atom, computed } from "nanostores";

const App = () => {
  /**
   * In this case we create a "custom" context limited to a request. This can be whatever
   * random memoized value we have.
   * Not sure if we'll go this route, but that'll require a first attempt at making it
   * work with a meta-framework (Next/Nuxt).
   *
   * I have a feeling that we'll need a middleware that would work out the machinery
   * of a meta-framework: create persistent req.id, initial render, `await allTasks`,
   * serialize context, deserialize context on client. So there's a high possibility
   * this all will be hidden from the end developer.
   */
  return (
    <NanostoresContext id="req_id">
      <User />
      <InputWrapper levels={5} />
    </NanostoresContext>
  );
};

/**
 * This is a "custom" atom. First time it's called with a "custom" context, it's marked
 * accordingly. After that you'll never be able to call it directly in module (this will
 * throw an error).
 */
const $user = atom({ username: "dkzlv" });

const User = () => {
  const user = useStore($user);
  return <p>Usename: {user.username}</p>;
};

/**
 * This looks to be a "custom" atom, but it in fact will be used as a local one,
 * because first time you use it (in `useStore`) you'll have a local context.
 */
const $inputValue = atom("");
/**
 * And even though `$inputValue` and `$user` are attached to different contexts,
 * this will work as expected! The resulting `computed` will also be a "local" atom.
 * Its local context has a parent context, which is the "custom" from above, so we
 * traverse the linked list until we find a fitting context.
 */
const $displayValue = computed(
  [$user, $inputValue],
  (user, input) => `${user.username} ++ ${input}`
);

const InputWrapper = ({ levels }: { levels: number }) => {
  return (
    /**
     * We hide the calls to `createLocalContext` and `createContext`, the surface is
     * the id property.
     */
    <NanostoresContext localId={`${levels}_id`}>
      <Input />
      {levels > 0 ? (
        // We can nest local contexts into each other!
        <InputWrapper levels={levels - 1} />
      ) : null}
    </NanostoresContext>
  );
};

const Input = () => {
  const inputValue = useStore($inputValue);
  const display = useStore($displayValue);
  const ctx = useCtx();

  return (
    <div>
      {display}
      <input
        value={inputValue}
        onChange={(e) => ctx($inputValue).set(e.currentTarget.value)}
      />
    </div>
  );
};
```

### DIFF
--- a/atom/index.js
+++ b/atom/index.js
@@ -74,12 +74,12 @@ export let atom = (initialValue, level = 0) => {
       listener(state.v)
       return unbind
     },
-    set value(newVal) {
-      getStoreState(this, $atom).v = newVal
-    },
     get value() {
       let state = getStoreState(this, $atom)
       return state.v
+    },
+    set value(newVal) {
+      getStoreState(this, $atom).v = newVal
     }
   }
 
--- a/context/index.d.ts
+++ b/context/index.d.ts
@@ -11,6 +11,11 @@ export type WithCtx = {
 export declare const globalContext: Context
 
 export function createContext(storeValues?: object): Context
+export function createLocalContext(
+  parentContext: Context,
+  id: string,
+  storeValues?: object
+): Context
 export function resetContext(context?: Context): void
 export function serializeContext(context: Context): string
 
--- a/context/index.js
+++ b/context/index.js
@@ -1,39 +1,51 @@
 let contexts = []
 
 // private object retains its id throught the lifetime
-let ctxTrait = {}
+let GLOBAL_CONTEXT_TRAIT = 1
+let GLOBAL_CUSTOM_CONTEXT_TRAIT = 2
+let LOCAL_CONTEXT_TRAIT = 3
 
 export function isContext(ctx) {
-  return ctx?._ === ctxTrait
+  return !!ctx?.t
 }
 
-function buildContext(storeStates = {}) {
+function buildContext(contextType = GLOBAL_CONTEXT_TRAIT, storeStates = {}) {
   let context = {
-    _: ctxTrait,
     // store copies
     copies: new Map(),
+    // nested contexts
+    nested: {},
     // listener [q]ueue
     q: [],
     // store states
-    states: storeStates
+    states: storeStates,
+    t: contextType
   }
   contexts.push(context)
   return context
 }
 
-let globalContextPolluted = false
 export const globalContext = buildContext()
 
 export function createContext(storeStates) {
-  globalContextPolluted = true
-  return buildContext(storeStates)
+  return buildContext(GLOBAL_CUSTOM_CONTEXT_TRAIT, storeStates)
 }
+
+export function createLocalContext(parentContext, id, storeStates) {
+  let cached = parentContext.nested[id]
+  if (cached) return cached
+
+  let ctx = buildContext(LOCAL_CONTEXT_TRAIT, storeStates)
+  parentContext.nested[id] = ctx
+  ctx.parent = parentContext
+  return ctx
+}
+
 export function resetContext(context) {
   if (context) {
     contexts = contexts.filter(c => context !== c)
   } else {
     contexts = []
-    globalContextPolluted = false
     delete globalContext.tasks
   }
 }
@@ -43,17 +55,10 @@ export function serializeContext(context) {
 
 // lazily initialize store state in context
 export function getStoreState(thisStore, originalStore) {
-  if (globalContextPolluted && thisStore.ctx === globalContext) {
-    if (process.env.NODE_ENV !== 'production') {
-      throw new Error(
-        `You can't mix global context and custom contexts, that is probably an error. ` +
-          'Please, pass the correct context into `withContext`.'
-      )
-    }
-    throw new Error('no global ctx')
-  }
+  let ctxFromTree = traverseContexts(originalStore, thisStore.ctx)
+  syncAtomTrait(originalStore, ctxFromTree)
 
-  let state = thisStore.ctx.states[originalStore.id]
+  let state = ctxFromTree.states[originalStore.id]
   if (!state) {
     state = {
       // [l]isteners [c]ount
@@ -63,7 +68,7 @@ export function getStoreState(thisStore, originalStore) {
       // [v]alue
       v: thisStore.iv
     }
-    thisStore.ctx.states[originalStore.id] = state
+    ctxFromTree.states[originalStore.id] = state
   }
   return state
 }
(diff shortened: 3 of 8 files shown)


dkzlv:
@AleksandrSl @ninaTorgunakova @euaaaio Yo guys and gals, added you to review this PR. Don't feel pressured to spend time on it, but your opinion may be valuable. If you're interested in implementation details, you can go ahead and dive into this and #228 PRs, but I don't expect so much, haha.
What I'm very much interested in is your perspective on context usage and edge cases I might have overlooked. All of you use (or used) nanostores extensively, and Eduard have contributed a lot into this repo and other nanostores-related projects, so you can share your perspective on lib development related to nanostores.

Share your thoughts. Можно по-русски, если вломак сидеть со словарём, хаха

euaaaio:
Will look into all of this carefully and write up my thoughts during the week.

<english mode off>
Спасибо, что позвал 🖤 
Пиары большие, важные и очень классные — мне нужно время, чтобы внимательно все изучить. Обязательно посмотрю их на неделе. Как минимум, постараюсь описать, как я это вижу в мире Vue.

Grawl:
А давайте не сокращать “context” до “ctx”

euaaaio:
> А давайте не сокращать “context” до “ctx”

Если ты говоришь про `useCtx()`, то, наверное, да. А если про код, то это оптимизация для экономии размера пакета.

Grawl:
Я про внешний интерфейс, конечно же. Внутри хоть обфусцируй :)

dkzlv:
Отвечу по-английски.

There are 2 reasons why I picked `ctx` over explicit `context`:

1. I think it would be better if we had a naming distinction between the framework's context and nanostore's one. It may seem like a dev won't really interact with nanostore's context at all, but in reality they probably will. So my initial idea was to mark everything in Nanostores as `ctx`, as we don't really have another name for it (and we shouldn't have, framerwork's context and nanostore's context serve the same purpose).
2. currently, there are a couple of places where a dev would be forced to somehow interact with contexts explicitly:
    - [in lifecycle hooks](https://github.com/nanostores/nanostores/pull/228/files#diff-233c9c805f9d87fde9e06ffd7bcb6901e0fbf502b1feb9074cb3525c79720cf4R73-R76). The plan is to provide the dev with a `ctx` function that's gonna wrap the atom and return any atom with the correct context
    - when you want to set the value in a component (see the `Input` component in the snippet in the issue text)
    
    The problem I saw is that if we use explicit names (e.g., `withContext` instead of a short `ctx`), it's gonna be so verbose that it's borderline unreadable. Like, if you need to access 5 different atoms inside an `onSet` hook, it's gonna be `const value1 = withContext($valueAtom1)` 5 times.

But, well, I don't have a strong opinion here, tbh.

Grawl:
I strongly appreciate that 'ctx' is widely used in a lot of code, especially in frameworks providing SSR.

But I have a principle: if my code becomes unreadable, verbose or too boilerplate, I should change how my code is designed.

Every single part of code should be self explanatory.

ninaTorgunakova (commented):
Honestly, I didn't get in which cases nesting local contexts using their own context id would be necessary, I personally think that it is a bit redundant in this API: mixing context Id and local context id doesn't seem minimalistic. Ideally, I would prefer the approach to use `NanostoresContext` only in one outer component without nesting it - to wrap the form (or particular input), and then completely forget about using `NanostoresContext` inside the inner components (with only call `useContext` where appropriate). Maybe I just don't want to think about where to pass `localId` and where to `id` prop, sort of mental overload in my opinion.

In other cases, the creation context seems a looks useful and cool feature.

Three groups of context in implementation, not only local and global but also 'custom', doesn't seem really easy to understand at first but if a developer shouldn't think about this classification and work differently with API, then it's good.

Regarding context/ctx - subjectively I would prefer context, +4 letters won't give too big difference, but no one will be confused for sure.

dkzlv:
@ninaTorgunakova 

> I didn't get in which cases nesting local contexts using their own context id would be necessary

Here's the thing. Putting all of the logic inside React actually has its own benefits. This way we tie the logic to an instance of the component, and the logic is kind of isolated. You can stop thinking about shared state problems, since you have a more or less strict guarantee component state isn't leaked and does not interact with the outer world.

(other than that having your logic in React sucks a lot)

But when we put everything into a JS module and nanostores, we lose this precious thing. And it in fact _is_ precious. Nesting local contexts is maybe a bit too specific, but it has some real cases. Simplest example I have is nested forms. Imagine having a product form from which you can also create/edit product categories. Those will be nested inside the product form, and from the perspective of React/Vue/etc. component tree will be something like 

```tsx
// This is the context for the entire form
<NanostoresContext localId="...">
  <ProductForm>
    {product.categories.map((c) => (
      // This is the context for a single nested form
      <NanostoresContext id={c.id}>
        <CategoryForm obj={c} />
      </NanostoresContext>
    ))}
  </ProductForm>
</NanostoresContext>;
```

Internally `CategoryForm` can have its own store with form data, validations, lifecycle hooks, etc., but since those atoms are local, you can put out as many `CategoryForm` instances in the tree without worrying too much about accidentally sharing the state.

> I would prefer the approach to use NanostoresContext only in one outer component without nesting it - to wrap the form (or particular input), and then completely forget about using NanostoresContext inside the inner components (with only call useContext where appropriate). Maybe I just don't want to think about where to pass localId and where to id prop, sort of mental overload in my opinion.

Full disclosure: this whole feature is not meant to be used unless you face a case that's not solvable without it! I aim at 100% compatibility with the prev nanostores version (I didn't change any tests significantly, and they all pass).

The overwhelming majority of the lib will remain the same. Global context is automatically created for you in the lib, custom context will be automatically created for you in the meta-framework integration code.

So, addressing your quote, even if you work a lot with complex forms, you probably won't create custom context yourself at all. As for the local context, well, maybe just a few times in your whole codebase. And you certainly won't need to dig into tech details of local vs custom vs global contexts, that's more of an internal design decisions and not part of the public dev-facing API.

(at least that's the intention)

ai:
I'm really afraid that nested context is too complex idea for this simple tasks.

My main concern is that you will need to look at all parents source to find that `useStore($inputStore)` works completely different here.

Right now, context is a simple thing, because we are sure that context is always the same in the app (at least for each user). But here we will not be able to trust `useStore()`, because some stores will be changed by local context, others won’t (and we would not be able to find out it without checking every parent).

Maybe we can pass input’s store in `props` to solve the problem without local context?

PR #238, opened 2023-09-03
Readme: Clarify `useStore`

tordans:
I miss read this section over and over again. I read is as "special helpers to re-render the component on any store changes:" where the next code section is this special helper. However, this was referring to `useStore` which is documented way below in the specific framework section.

This should clarify this issue and make sure users look for "`useStore`" in the doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,8 +139,9 @@ $counter.set($counter.get() + 1)
 ```
 
 `store.subscribe(cb)` and `store.listen(cb)` can be used to subscribe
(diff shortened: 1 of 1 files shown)


euaaaio:
I don't like that an important description of why we have integrations to all frameworks is in the description of the store methods.

Maybe it would be better to complete it and move it to [the Integrations section](https://github.com/nanostores/nanostores/blob/main/README.md#integration)?

ai:
It is just links. Links are nice.

PR #239, opened 2023-09-04
Allow `undefined` in `setKey`

gabrielsoldani:
To allow clearing a key when passing `undefined` as the value, we should add `undefined` to the parameter type signature.


gabrielsoldani:
Ran into a typing error trying to make a defensive `delete` wrapper :)

ai:
Should we allow clean key if was not marked as optional in types?

ai:
I am closing PR since it could set `undefined` to the key which was not allows to be `undefined`

PR #244, opened 2023-09-13
docs: update expired link to Svelte's store contract

gismya:
Reading the docs I noticed that the link to the Svelte documentation describing its store contract was no longer valid, this updates the link.


ai:
Thanks!

PR #249, opened 2023-10-02
Add listenerLevel argument for listen methods to Atom types

### LINKED ISSUE #248: Untyped methods and arguments
Hello!

I'm implementing nanostores in a project and noticed that a few methods in the ReadableAtom are not reflected in the types. Specifically `notify`, and the arguments for `level` in the initializing function, and the arguments for `listenerLevel` in the `subscribe` and `listen` methods. Are they supposed to be internal or should they be included in the types? If so I can create a PR for that.

gismya:
Resolves #248

### DIFF
--- a/atom/index.d.ts
+++ b/atom/index.d.ts
@@ -47,9 +47,10 @@ export interface ReadableAtom<Value = any> {
    * immediately.
    *
    * @param listener Callback with store value.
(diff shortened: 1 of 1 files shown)


ai:
Got it. No, this argument is private as well. You should not use it, if you are not creating something very low-level.

Types for that argument in VS Code will only confuse casual developers.

PR #250, opened 2023-10-14
fix: missed recalculation of `computed` with `deepMap` inside it

euaaaio:
The problem was the lack of immutability as it works in `setKey()` of `map()`.


ai:
Thanks. Released in  0.9.4

PR #254, opened 2023-10-21
simpler synchronous only autosubscribe behavior to computed

btakita:
This is a simpler implementation than https://github.com/nanostores/nanostores/pull/221

The autosubscribe `computed` does not take an argument.

```ts
let $a = atom(1)
let $b = atom(2)
computed(() => $a() + $b())
```

size-limit:

- Atom: + 20B
- All: + 119B


ai:
1. I still don’t like that `$store()` API is only for `computed()`. Looks like making API more complex.
2. Nano Stores API design is to keep thing simple. Less auto-magic (except diamond problem, but it was a fix, not a design choice) and more straightforward. The core idea: “I know how it is written inside” just by looking on the API.

btakita:
I ended up creating my own library called [rmemo](https://github.com/ctx-core/rmemo). It was originally forked from VanJS & uses some of the concepts in Nanostores. It is a reactive foundation for several other libraries. I'm happy with the result. Good luck with this project!

PR #266, opened 2024-01-25
feat!: Update computed values before other listeners

gismya:
An idea that sprouted from the discussion here: https://github.com/orgs/nanostores/discussions/265.

When using computed values in listeners of the values they depend on the computed have stale values. This PR changes the priority of the value updates for computed values to be lower than the original atom without changing the priority of the listeners of the computed values.

(Sidenote: how are the functions included in the "All" size check decided? It's not including things like deepMap and listenKeys?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ A tiny state manager for **React**, **React Native**, **Preact**, **Vue**,
 **Svelte**, **Solid**, **Lit**, **Angular**, and vanilla JS.
 It uses **many atomic stores** and direct manipulation.
 
-* **Small.** Between 278 and 804 bytes (minified and brotlied).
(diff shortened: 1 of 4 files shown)


ai:
Can you rebase the changes to `next` branch since it could be a major changes?

gismya:
> Can you rebase the changes to `next` branch since it could be a major changes?

Done

> There is not a strict system. Maybe we should rename it. I made a note to do it.

Went from a note to a commit very quickly :)

ai:
Thanks!

gismya:
This seems to now be gone from the next branch. Did it get lost while rebasing?

ai:
Fuck. Fixed https://github.com/nanostores/nanostores/commit/c6765a1be4525c3fa1f3840785caf272152232bc

Issue #197, opened 2023-05-25
Releases on github


dartess:
Is it possible to use github's `releases`? Most of the packages I know have a "Releases" page with changelogs. This allows you to subscribe to new releases using the native Watch mechanism.

dartess:
Unfortunately no.

> You can watch for tags :)

I don't seem to be able to get notifications for new tags.

I meant that you can subscribe to releases and receive notifications.

<img width="336" alt="image" src="https://github.com/nanostores/nanostores/assets/15047511/b78bd51f-b112-4c5c-9cef-9ee4b33aaaa1">

dartess:
It seems releases are made from tags, and you are already actively using tags.
It would be great to publish releases, and instead of changelogs, put a link to the `CHANGELOG.md` file.

dartess:
I don't have relevant experience.

euaaaio:
We don’t need to invent the new one — https://github.com/yyx990803/release-tag

azat-io:
+1 for releases. I use [GitHub release notify telegram bot](https://github.com/pyatyispyatil/github-releases-notify-bot) to get notification of new releases. This thing works just for GitHub releases.

[Release tag](https://github.com/yyx990803/release-tag) is the easiest way. But all the release messages will look something like [this](https://github.com/vuejs/vitepress/releases).

Personally, I prefer other tools, but they all work with [conventional commits](https://conventionalcommits.org/).

ai:
Done https://github.com/nanostores/nanostores/blob/next/.github/workflows/release.yml

PR #267, opened 2024-02-08
Releases on github #197

edloidas:
- Added release task after checking for push tags (release will be skipped if no notes are found for a tag in the CHANGELOG.md file).
- Added `contents: write` permissions to the workflow (required to create a release).
- Removed `--frozen-lockfile` as it is [automatically used](https://pnpm.io/next/cli/install#--frozen-lockfile) on CI. 
- Updated versions of actions used in workflow.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,44 @@
+name: Release
+on:
+  push:
+    tags:
+      - '*'
+permissions:
+  contents: write
+jobs:
+  release:
+    name: Release On Tag
+    if: startsWith(github.ref, 'refs/tags/')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v4
+      - name: Extract the changelog
+        id: changelog
+        run: |
+          TAG_NAME=${GITHUB_REF/refs\/tags\//}
+          READ_SECTION=false
+          CHANGELOG_CONTENT=""
+          while IFS= read -r line; do
+            if [[ "$line" =~ ^#+\ +(.*) ]]; then
+              if [[ "${BASH_REMATCH[1]}" == "$TAG_NAME" ]]; then
+                READ_SECTION=true
+              elif [[ "$READ_SECTION" == true ]]; then
+                break
+              fi
+            elif [[ "$READ_SECTION" == true ]]; then
+              CHANGELOG_CONTENT+="$line"$'\n'
+            fi
+          done < "CHANGELOG.md"
+          CHANGELOG_CONTENT=$(echo "$CHANGELOG_CONTENT" | awk '/./ {$1=$1;print}')
+          echo "changelog_content<<EOF" >> $GITHUB_OUTPUT
+          echo "$CHANGELOG_CONTENT" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+      - name: Create the release
+        if: steps.changelog.outputs.changelog_content != ''
+        uses: softprops/action-gh-release@v1
+        with:
+          name: ${{ github.ref_name }}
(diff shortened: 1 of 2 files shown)


edloidas:
There are several options for creating a release, but I find using the same workflow file to work the best. You can see other options, how release notes will look like and some reasoning in my [demo repo](https://github.com/edloidas/automation).

.github/workflows/test-and-release.yml
```diff
@@ -0,0 +1,111 @@
...
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: pnpm
+      - name: Install dependencies
+        run: pnpm install --ignore-scripts
+      - name: Run benchmark
+        run: node ./test/benchmark.js
+  release:
```
ai, line 75: Should we move it to another workflow since we do not need to run it on PRs

.github/workflows/test-and-release.yml
```diff
@@ -0,0 +1,111 @@
...
+    name: Release On Tag
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
+    needs: [full, short, benchmark]
```
ai, line 78: Let’s create a release on new tag without waiting for tests.

Anyway tag was created.

.github/workflows/test-and-release.yml
```diff
@@ -0,0 +1,111 @@
...
+                break
+              fi
+            elif [[ "$READ_SECTION" == true ]]; then
+              CHANGELOG_CONTENT+="$line"$'\n'
+            fi
+          done < "CHANGELOG.md"
+          CHANGELOG_CONTENT=$(echo "$CHANGELOG_CONTENT" | awk '/./ {$1=$1;print}')
+          echo "changelog_content<<EOF" >> $GITHUB_OUTPUT
+          echo "$CHANGELOG_CONTENT" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
```
ai, line 103: Can you show an example of description?

ai: Found in https://github.com/edloidas/automation

.github/workflows/test-and-release.yml
edloidas: This will simplify the workflow. Given the previous comment, we'll have a total of three workflow files: test, benchmark (won't trigger on PR), release (won't depend on tests). Is this correct?

.github/workflows/test-and-release.yml
edloidas: Sure. It is in the demo repo I mentioned above. Check out [CHANGELOG.md](https://github.com/edloidas/automation/blob/main/CHANGELOG.md) and [releases](https://github.com/edloidas/automation/releases) with notes.

.github/workflows/test-and-release.yml
ai: I think we should have 2 workflow files: `test.yml` and `release.yml`

edloidas: Ah yes, sorry, thought the previous conversation was pointing to the benchmark.

ai: We can keep `benchmark` in `test.yml` as we had before (since we need to run it on PR)

edloidas:
Moved the release to a separate workflow that doesn't depend on the test jobs. Now it looks even simpler, as there is no need for some checks. 
Test workflow almost intact, but I kept the actions upgrade.

ai (approved):

ai:
Everything looks good. I will merge it after the lunch.

ai:
Thanks. Great workflow. I am [going to use it](https://github.com/ai/boilerplates/commit/72b971304a0e427a8f57fb4c7b651c67207e1132) in all my open source projects.

PR #273, opened 2024-02-20
Demonstrate memory leak

hypeJunction:
I have been debugging an issue with orphan subscriptions being left behind, and it led me to the source code. Could you please confirm whether this is an intended behaviour? It feels like there is a bug in there.


ai:
Yes, it is very strange

ai:
We didn’t have leak. I fixed the test e52eab4

`computed` wait for a second for new subscription and only then unsubscribes. It is useful in React/Svelte apps on page routing (when everything was unsubscribed and re-subscribed again in 100ms).

PR #277, opened 2024-02-27
Add notify() store method to public types

gismya:
Suggestion to re-add `notify()` to the public types following discussion in https://github.com/orgs/nanostores/discussions/276


ai:
Thanks

PR #280, opened 2024-03-06
feat!: does not run queued listeners after they are unsubscribed

gismya:
Changes the behaviour of unsubscribe to also remove listeners that have already been queued.

It's a breaking change.

I remade the tests "does not mutate listeners while change event". Can't really work out what it was testing, and if/how I should replace it with something else.


ai:
Thanks

PR #283, opened 2024-03-26
Add missing `Task` type in `batch` types 

mxck:
Allow to use `batched` with `task`.

Little example:
```ts
const someStore1 = atom('test');
const someStore2 = atom('test');

const batchedStore = batched([someStore1, someStore2], () => {
  return task(async () => {
    return 'task';
  });
});

batchedStore.subscribe(() => {});

setTimeout(() => {
  // Throws type error
  const value: string = batchedStore.get(); // Expected type 'string', got 'Task<string>'
});
```

### DIFF
--- a/computed/index.d.ts
+++ b/computed/index.d.ts
@@ -53,7 +53,7 @@ export const computed: Computed
 interface Batched {
   <Value extends any, OriginStore extends Store>(
     stores: OriginStore,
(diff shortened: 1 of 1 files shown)


mxck:
Oh sorry, this code doesn't work in runtime :\

ai:
Thanks. I am going to release it during the week.

PR #285, opened 2024-04-01
fix: missed passing old value to `onNotify` callback

euaaaio:


ai:
Thanks. Release in 0.10.1.

PR #290, opened 2024-04-05
fix(deepMap): incorrect previous value when listening deep store

euaaaio:
To use `structuredClone` as part of the core, we would have to rewrite `notify`, which would add about 20 bytes to the `atom`, just to support `deepMap` mutability.

I think until we make `notify` a public API, it's better to use `structuredClone` only inside `deepMap`.


ai:
Thanks. Released in 0.10.3.

PR #300, opened 2024-05-03
Fix stale computed values

### LINKED ISSUE #299: computed store can return stale value
Here's a test case to illustrate:

```ts
test('stale computed', () => {
  let $atom = atom(1)
  let values: (number | string)[] = []
  let $computed1 = computed($atom, () => {values.push($computed2.get())})
  let $computed2 = computed($atom, value => value * 2)
  $computed1.get()
  $atom.set(2)
  deepStrictEqual(values, [2, 4])
})
```

`$computed2.get()` inside the `$computed1` callback should return `2` and then `4`. But it actually returns `2` and then `2`.

russelldavis:
Fixes #299

This fixes stale values in computed stores by checking if dependencies have changed any time `get()` is called. It does have a perf impact, though I added an `epoch` optimization to help with that. And correctness should take precedence over performance (no other state library I'm aware of has this stale value problem). The benefits are:

- computed stores never return stale values
- this also solves the diamond problem, so the previous logic that was added for that is no longer needed, which means:
  - simpler logic
  - no need to store and compute a level (`l`) for each atom
  - bundle size reduced (min: 286B to 235B, max:  818B to 769B)
  - better perf in certain areas -- processing the `listenerQueue` is now O(n) instead of O(n^2)

Another potential benefit -- this change means that when calling `get()` on an atom with 0 subscribers, we no longer have to [add a dummy subscriber](https://github.com/nanostores/nanostores/blob/7fb5cd871fa00f484ebc866fdf330b56cab2aa20/atom/index.js#L9-L11). I'm holding off on making that change, because it would be a breaking change -- some code could be relying on the side effect of `onMount` firing. But it could be considered for a future major version -- IMO it would be cleaner. Right now, if you call `get()` on a computed store, it will never unsubscribe, and the value will be needlessly recomputed when the dependencies change, even if the only listener is the dummy listener.

### DIFF
--- a/atom/index.js
+++ b/atom/index.js
@@ -2,7 +2,7 @@ import { clean } from '../clean-stores/index.js'
 
 let listenerQueue = []
 
-export let atom = (initialValue, level) => {
+export let atom = (initialValue) => {
   let listeners = []
   let $atom = {
     get() {
@@ -11,25 +11,23 @@ export let atom = (initialValue, level) => {
       }
       return $atom.value
     },
-    l: level || 0,
     lc: 0,
-    listen(listener, listenerLevel) {
-      $atom.lc = listeners.push(listener, listenerLevel || $atom.l) / 2
+    listen(listener) {
+      $atom.lc = listeners.push(listener)
 
       return () => {
         let index = listeners.indexOf(listener)
         if (~index) {
-          listeners.splice(index, 2)
+          listeners.splice(index, 1)
           if (!--$atom.lc) $atom.off()
         }
       }
     },
     notify(oldValue, changedKey) {
       let runListenerQueue = !listenerQueue.length
-      for (let i = 0; i < listeners.length; i += 2) {
+      for (let listener of listeners) {
         listenerQueue.push(
-          listeners[i],
-          listeners[i + 1],
+          listener,
           $atom.value,
           oldValue,
           changedKey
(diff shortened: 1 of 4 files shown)


ai:
> this change means that when calling get() on an atom with 0 subscribers, we no longer have to [add a dummy subscriber](https://github.com/nanostores/nanostores/blob/7fb5cd871fa00f484ebc866fdf330b56cab2aa20/atom/index.js#L9-L11)

Is It possible to return this behaviour?

I use it a lot on my CRDT stores. The store value before `onMount` is just undefined.

> Right now, if you call get() on a computed store, it will never unsubscribe

Seems like it is a bug. The dummy subscriber for listener-less `get()` should be removed after.

russelldavis:
> Is It possible to return this behaviour?

I didn't change the behavior in this PR since it would be a breaking change. It's just a suggestion for a possible future change that would be possible after this PR. We can ignore it for now and discuss it as a separate issue if this PR gets merged.

ai:
In this case everything looks amazing. I will merge it to `next` and release in next major release (I am afraid on invisible breaking changes).

ai:
I merge to `next` branch and it broke the tests :(

Can you find the reason? https://github.com/nanostores/nanostores/actions/runs/8958545657/job/24602887313

russelldavis:
Fixed at #301

PR #301, opened 2024-05-05
Fix merge of #300

russelldavis:
The new code in #380 needed to be updated to reflect the new lower number of items in listenerQueue per listener.


ai:
Thanks! It was my mistake of merging before the CI finished after changing the base.

PR #302, opened 2024-05-06
Fix issues with removing listeners from the queue

russelldavis:
After switching to the `next` branch I discovered several issues introduced by #280:

- It was passing `index` instead of `queueIndex` to `listenerQueue.splice`. In addition to being the wrong index, this can put `listenerQueue` into a really weird/bad state when `index` is not divisible by 4 (since `listenerQueue` uses chunks of 4 items per listener).
- It was only removing the first listener it found in the queue, so it wouldn't work properly if the same listener was in the queue multiple times (e.g., when calling `set` on the same store multiple times inside a listener)
- It was removing items from the front of the queue, even if those items had already been processed. This would cause the next iteration of the queue processing loop to skip unprocessed items (because everything in the array gets shifted when calling `splice`)
- It was searching all items of `listenerQueue` for `listener`, but as mentioned above, only every 4th item is a valid position to search in. This meant that if a listener function was set as the value of a store, it would be incorrectly found as a listener, and `listenerQueue` would end up in a bad state. This is an unlikely scenario to happen in the real world, but easy enough to fix.

I added tests for all of those and fixed the issues.

cc @gismya

### DIFF
--- a/atom/index.js
+++ b/atom/index.js
@@ -1,6 +1,8 @@
 import { clean } from '../clean-stores/index.js'
 
 let listenerQueue = []
+let lqIndex = 0
+const QUEUE_ITEMS_PER_LISTENER = 4
 export let epoch = 0
 
 export let atom = (initialValue) => {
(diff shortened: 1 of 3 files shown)


gismya:
A superior solution. While the first issue was an embarrassing typo, I didn't consider the other issues and went for an overly simplistic solution. I love the clarity of the tests.

ai:
Thanks!

PR #303, opened 2024-05-08
Improve computed perf by setting currentEpoch after set

russelldavis:
The benchmark for `computed` on my machine increased from 3,815 ops/sec to 5,193 ops/sec


ai:
♥️ amazing

PR #305, opened 2024-05-16
Don't mutate deepMap values

russelldavis:
Previously, calling `setKey` on a $deepMap would mutate the value in-place. This can cause various problems, for example #250 and #290. Those were both solved with workarounds, however, this solves it in a better way, by not mutating the deepMap in the first place.

Some advantages of this approach:
* No need for structuredClone
* The oldValue passed to listeners is the same object reference as the value that was previously passed. This is how immutable stores usually work, so it will align better with expectations, cause fewer issues, and allow for new use cases, like the upcoming batching PR which needs that behavior.
* Copies were already being made at each nesting level of the key being set, they just weren't being used (so maybe this was the original intent?). Either way, that means there's no new work being done, and perf will actually improve since we no longer need to call structuredClone on oldValue.

As part of this change, when a key path extends past the end of an array the result will be a sparse array rather than filling it with undefined. It would have required extra code (and slower perf) to keep the old behavior, and I don't see a good reason for it. The new behavior aligns with what JS does natively, and with other stores like SolidJS. But let me know if that's a problem.

### DIFF
--- a/deep-map/index.d.ts
+++ b/deep-map/index.d.ts
@@ -32,7 +32,9 @@ export type DeepMapStore<T extends BaseDeepMap> = {
   notify(oldValue?: T, changedKey?: AllPaths<T>): void
 
   /**
-   * Change key in store value.
+   * Change key in store value. Copies are made at each level of `key` so that
+   * the old value is not mutated (but it does not do a full deep copy --
+   * references to objects will still be shared between the old and new value).
    *
    * ```js
    * $settings.setKey('visuals.theme', 'dark')
--- a/deep-map/index.js
+++ b/deep-map/index.js
@@ -6,14 +6,9 @@ export { getPath, setByKey, setPath } from './path.js'
 export function deepMap(initial = {}) {
   let $deepMap = atom(initial)
   $deepMap.setKey = (key, value) => {
-    let oldValue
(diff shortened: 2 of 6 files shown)


ai:
Can you also explain a little more how do you do it? (Sorry, I am full in another project right now for deep review)

russelldavis:
It all happens in `setByKey`. It uses a copy of each object/array at each level of the key path, so that nothing in the original object tree is ever mutated. It's like making a deep copy, but only on the parts of the object tree that will be mutated (the rest is a shallow copy).

ai:
So if we change `a.b` key, it means what will be mutated only?

```
{ // mutates
  a: { // mutates
    b: 'new value', // mutates
    c: {} // copies
  },
  d: {} // copies
}
```

Nice, I like the idea.

russelldavis:
Yeah, so in your example, after you call `$store.setKey("a.b", "foo")`, `$store.value`, `$store.value.a` and `$store.value.a.b` will be new objects, while `$store.value.a.c` and `$store.value.d` will remain as the same unmodified objects.

russelldavis:
Thanks for merging. cc @dkzlv @euaaaio for awareness

PR #306, opened 2024-05-18
Add `batch` function for batching writes

russelldavis:
Supersedes #298. See the description of that PR for the overview. Changes in this version:

* Based on `next` branch instead of `main`
* Fixes stale values and listeners being called multiple times when batching
* Removes `notify` from the public types, adds overridable `isEqual` function as alternative
* Fixes listenKeys/subscribeKeys to work with `batch`
* Adds documentation

There are more details about each of those bullet points in the commit messages. It's probably best to start by looking at the commits individually rather than all the changes together.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ A tiny state manager for **React**, **React Native**, **Preact**, **Vue**,
 **Svelte**, **Solid**, **Lit**, **Angular**, and vanilla JS.
 It uses **many atomic stores** and direct manipulation.
 
-* **Small.** Between 286 and 818 bytes (minified and brotlied).
+* **Small.** Between 287 and 818 bytes (minified and brotlied).
   Zero dependencies. It uses [Size Limit] to control size.
 * **Fast.** With small atomic and derived stores, you do not need to call
   the selector function for all components on every store change.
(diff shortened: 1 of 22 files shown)


README.md
```diff
@@ -356,39 +371,72 @@ export const $user = computed($userId, userId => task(async () => {
 }))
 ```
 
-By default, `computed` stores update _each_ time any of their dependencies
-gets updated. If you are fine with waiting until the end of a tick, you can
-use `batched`. The only difference with `computed` is that it will wait until
-the end of a tick to update itself.
+### Batching
```
ai, line 374:
```suggestion
### Batching

```

atom/index.js
```diff
@@ -14,9 +32,19 @@ export let atom = (initialValue) => {
       }
       return $atom.value
     },
+    isEqual: Object.is,
     lc: 0,
-    listen(listener) {
+    listen(_listener) {
```
ai, line 37: `_var` is not cool, let’s find a better name for next callback.

ai:
Can I ask you to rebase the PR? Then we are ready to merge it and release it (sorry, for the delay which lead to conflict).

atom/index.js
```diff
@@ -11,69 +32,56 @@ export let atom = (initialValue, level) => {
       }
       return $atom.value
     },
-    l: level || 0,
+    isEqual: Object.is,
```
ai, line 35: By moving to `eq` or `is` we can save some size. This API is very rare, so it is OK to use some cryptic name.

PR #309, opened 2024-05-21
fix(types): add `undefined` to map.setKey type

### LINKED ISSUE #308: TypeScript error with `map.setKey` when passing `undefined` to un-set a key
To un-set a `map`'s key, we pass `undefined` to the `setKey` method:

```js
const $store = map({});
$store.setKey('some_key', undefined) // OK
```

When using TypeScript, this errors when the value itself is not typed as `SomeType | undefined`:

```ts
const $store = map<Record<string, string>>({});
$store.setKey('some_key', undefined) // Argument of type 'undefined' is not assignable to parameter of type 'string'.ts(2345)
```

psychedelicious:
Closes #308

### DIFF
--- a/map/index.d.ts
+++ b/map/index.d.ts
@@ -89,7 +89,7 @@ export interface MapStore<Value extends object = any>
    */
   setKey<Key extends AllKeys<Value>>(
     key: Key,
-    value: Get<Value, Key> | Value[Key]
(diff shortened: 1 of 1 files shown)


ai:
As I described in #https://github.com/nanostores/nanostores/issues/308 I can’t accept it, because it will allow removing required key

PR #311, opened 2024-06-01
add downloads image

cwtuan:
Thank you for creating and maintaining the nanostores package. I've greatly benefited from its features and simplicity in managing state in my applications.

In this pull request, I've added a downloads trend image to the README file. This visual representation shows the increase in popularity and adoption of the nanostores package over recent years. New users often gauge the reliability and stability of a package based on its popularity. The trend image provides a quick, visual confirmation that nanostores is a trusted and growing library.

Preview:
image

Thank you for your time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,11 @@ It uses **many atomic stores** and direct manipulation.
   in the chunk.
 * Designed to move logic from components to stores.
 * Good **TypeScript** support.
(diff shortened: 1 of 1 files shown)


ai:
Thanks for thinking about our marketing.

It is powerful point.

But I think it is a little dangerous. Our industry is really suffering from hype-based development. I even have a whole talk about [that problem](https://slides.com/ai/frontend-is-dying).

So I prefer to not use this argument. We should not promote choosing tool according to downloads number.

PR #313, opened 2024-06-21
Decrease size of computed function

Minhir:
Stripped out 14B.   The `currentRunId` is not really needed for tracking the stale state, since we already have `previousArgs` and `args`, which are sufficient.

I also covered staled behavior with test.

### DIFF
--- a/computed/index.js
+++ b/computed/index.js
@@ -5,22 +5,17 @@ let computedStore = (stores, cb, batched) => {
   if (!Array.isArray(stores)) stores = [stores]
 
   let previousArgs
-  let currentRunId = 0
   let currentEpoch
   let set = () => {
     if (currentEpoch === epoch) return
     currentEpoch = epoch
     let args = stores.map($store => $store.get())
-    if (
-      previousArgs === undefined ||
-      args.some((arg, i) => arg !== previousArgs[i])
-    ) {
-      let runId = ++currentRunId
+    if (!previousArgs || args.some((arg, i) => arg !== previousArgs[i])) {
       previousArgs = args
       let value = cb(...args)
       if (value && value.then && value.t) {
         value.then(asyncValue => {
-          if (runId === currentRunId) {
+          if (previousArgs === args) {
             // Prevent a stale set
             $computed.set(asyncValue)
           }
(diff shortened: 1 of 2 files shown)


ai:
Nice! I will check it a few days later, when I will end my travels.

ai:
It was added in https://github.com/nanostores/nanostores/pull/229 by @btakita.

@btakita what is use case for `currentRunId`?

Minhir:
Sure, it would be nice if @btakita has time to validate my PR as well.

I can give my understanding of `currentRunId`. The problem here is that we can have multiple simultaneously running async tasks with different execution time. Let's say we have `task1` (1 sec), `task2` (3 sec) and `task3` (2 sec) started one after another (the time in parentheses is how long it takes for the task to be resolved). In such case we will see first a result from `task1` then the result from `task3` and the `task2` result will be the latest. But it looks very confusing from user's point of view. And  `currentRunId` could be used here as a unique `id` of the latest running task. So the code `runId === currentRunId` is nothing more than just checking if the current task is the latest one. You can try to remove `let runId = ++currentRunId` line and run the new test I added in this PR. It will fail. 

With all that said above, we can look at my change. For each new task, we are running this line `let args = stores.map($store => $store.get())` which creates a new `args` array (and a unique reference). Also, we already have `previousArgs` for simple memoization. From semantic point of view, we can rename `previousArgs` to `latestArgs` to make it more precise. And now we're ready to make an observation: we have a reference to `previousArgs` and we also have a reference to `args` which is unique for each task. We can use that reference as a task `id`.

btakita:
LGTM. I recall that `currentRunId` was used to make the tests pass at some point. If the tests still pass after removing `currentRunId`. All should be ok.

ai:
I rebased it to `next`, but seems like we already removed `currentRunId` there. Can you check?

Minhir:
@ai not really. Epoch is something similar, but it's global. So we still need some flag for defining of stale state. 
I've rebased everything. My test still will fail if you remove `currentRunId` or my changes. 

BTW, do you always accept changes in next branch? In such case, I propose to add Contributing part into README to make more clear :smiley:

ai:
> BTW, do you always accept changes in next branch? In such case, I propose to add Contributing part into README to make more clear

In most of cases we accept to `main`, but now we are preparing `next` branch.

PR #320, opened 2024-07-10
Add subscribe() tests

hypeJunction:
Adds some tests to inspect the behaviour of subscribe(), namely to ensure that the listener is actually receiving an old value of the atom

### DIFF
--- a/atom/index.test.ts
+++ b/atom/index.test.ts
@@ -125,6 +125,7 @@ test('supports complicated case of last unsubscribing', () => {
 
 test('supports the same listeners', () => {
   let events: (string | undefined)[] = []
+
   function listener(value: string | undefined): void {
     events.push(value)
   }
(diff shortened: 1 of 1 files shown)


hypeJunction:
I am seeing weird behavior with `subscribe()`. For some reason, I am not getting the expected `oldValue` as second argument in the listener. Maybe you could clarify if I am doing something wrong, or if there is an issue in code.

Also, it would be great if you documented how to contribute to this package. I couldn't make sense of your linting setup.

atom/index.test.ts
```diff
@@ -1,8 +1,8 @@
-import FakeTimers, { type InstalledClock } from '@sinonjs/fake-timers'
-import { test } from 'uvu'
-import { equal, is } from 'uvu/assert'
+import FakeTimers, {type InstalledClock} from '@sinonjs/fake-timers'
```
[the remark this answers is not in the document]
hypeJunction: Ah, I was trying to make eslint do the linting, but couldn't. Let me try prettier.

atom/index.test.ts
```diff
@@ -317,4 +328,34 @@ test('prevents notifying when new value is referentially equal to old one', () =
   clock.runAll()
 })
 
+test('notifies the subscribed listener with current and old values for a store that had an initial value', () => {
```
ai, line 331: Can you resolve conflict (I also changed branch to `next` where we work on the next version).

Without it I can’t run CI.

atom/index.test.ts
```diff
@@ -308,3 +309,35 @@ test('prevents notifying when new value is referentially equal to old one', () =
...
+  })
+
+  $store.set(1)
+  equal(events, [undefined, 1])
+
+  $store.set(2)
+  equal(events, [undefined, 1, 1, 2])
+})
+
+test.run()
```
ai, line 343: We don’t need to call `test.run` since we moved from `uvu` to `node:test`

hypeJunction:
Let me start from scratch and do this on the next branch. I am confused what's happening.

hypeJunction:
My upstream was way behind :man_facepalming:

PR #321, opened 2024-07-10
Add more subscribe() tests

hypeJunction:
Not sure why these fail, while the ones before testing the oldValue aren't.

### DIFF
--- a/atom/index.test.ts
+++ b/atom/index.test.ts
@@ -424,3 +424,32 @@ test('can use previous value in subscribers', () => {
   unbind()
   clock.runAll()
 })
+test('notifies the subscribed listener with current and old values for a store that had an initial value', () => {
+  let events: (number | undefined)[] = []
+  let $store = atom<number>(1)
+
+  $store.subscribe((value, oldValue?: number) => {
+    events.push(oldValue, value)
+  })
+
+  $store.set(2)
+  equal(events, [1, 2])
+
+  $store.set(3)
+  equal(events, [1, 2, 2, 3])
+})
+
+test('notifies the subscribed listener with current and old values for a store that had no initial value', () => {
+  let events: (number | undefined)[] = []
+  let $store = atom<number | undefined>(undefined)
+
+  $store.subscribe((value, oldValue?: number) => {
+    events.push(oldValue, value)
+  })
+
+  $store.set(1)
+  equal(events, [undefined, 1])
+
+  $store.set(2)
+  equal(events, [undefined, 1, 1, 2])
+})
--- a/atom/index.test.ts
+++ b/atom/index.test.ts
@@ -433,11 +433,13 @@ test('notifies the subscribed listener with current and old values for a store t
     events.push(oldValue, value)
   })
 
+  deepStrictEqual(events, [undefined, 1])
+
   $store.set(2)
-  equal(events, [1, 2])
+  deepStrictEqual(events, [undefined, 1, 1, 2])
 
   $store.set(3)
-  equal(events, [1, 2, 2, 3])
+  deepStrictEqual(events, [undefined, 1, 1, 2, 2, 3])
 })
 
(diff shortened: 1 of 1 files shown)


atom/index.test.ts
```diff
@@ -424,3 +424,32 @@ test('can use previous value in subscribers', () => {
   unbind()
   clock.runAll()
 })
+test('notifies the subscribed listener with current and old values for a store that had an initial value', () => {
```
ai, line 427:
```suggestion

test('notifies the subscribed listener with current and old values for a store that had an initial value', () => {
```

atom/index.test.ts
```diff
@@ -424,3 +424,32 @@ test('can use previous value in subscribers', () => {
...
+  let events: (number | undefined)[] = []
+  let $store = atom<number>(1)
+
+  $store.subscribe((value, oldValue?: number) => {
```
ai, line 432: You need to have here `listen`.

`listen` will be called on changes, when `subscribe` is called on changes and on initial state (where `oldValue` will be `undefined` of course).

hypeJunction: I don't understand. Why would it be `undefined` if the initial store value is `1`?

hypeJunction: Ah, I think I am starting to understand. At the time the subscription is established there is no change in value and the handler just receives the current value.

ai: Maybe we have issue. Let’s change test a little:

1. `events.push(oldValue, value)` → `events.push([oldValue, value])` so we will see pairs.
2. `equal(events, [1, 2])` → `equal(events, [[1, undefined], [1, 2]])`
3. The same with other tests and checks

atom/index.test.ts
```diff
@@ -424,3 +424,37 @@ test('can use previous value in subscribers', () => {
...
+
+test('notifies the subscribed listener with current and old values for a store that had an initial value', () => {
+  let events: (number | undefined)[] = []
+  let $store = atom<number>(1)
+
+  $store.subscribe((value, oldValue?: number) => {
+    events.push(oldValue, value)
+  })
+
+  deepStrictEqual(events, [undefined, 1])
```
ai, line 436: No, the `value` here should be `1` since store is already initialized with `1` on the moment when we call `subscribe`

hypeJunction:
@ai I think there is a bigger question of how would we know if the handler is being called upon initialization or if the value has actually changes. In my use case, I need the current value, and I need to know if that value has actually changes, so that I can make the decision about the course of action. Currently, I would need to introduce some local state to keep track of that.

Perhaps it would make sense to make value = oldValue, that way you would at least be able to tell that this is an initial cal to the handler.

gismya:
> I need the current value, and I need to know if that value has actually changed

If the value hasn't changed the subscriber callback won't be called, https://github.com/hypeJunction/nanostores/blob/89be3938f13a46670307fc24a09ea5d94fcb4f48/atom/index.js#L65-L69

hypeJunction:
I understand. My problem is the opposite. When the callback is called I don't know if it's because value has changed from `undefined` or if it's an initial call.

ai:
Just use `store.listen(cb)`. It doesn't have initialisation calls, only on value changes

hypeJunction:
But I need the current value. 

I am watching a connection status when my component mounts. The connection could have been established already or it might be established in the future. I think I need subscribe rather than listen.

hypeJunction:
I think I will just switch to using events for this, otherwise it gets all very convoluted

ai:
```js
function cb (value, initialization) {}
cb(store.get(), true)
store.listen(value => {
  cb(value, false)
})
```

hypeJunction:
Elegant, thanks. I have been overcomplicating the problem

ai:
I tested behavior and everything works as expected

PR #322, opened 2024-07-12
Improve types

psd-coder:
This PR improves types for two things:
1. At the moment `atom.value` always includes `undefined` in its type, despite whether initial value was provided or not. So our test usually sprinkled with optional chaining operator (`?.`). This PR changes it and if initial value was specified in atom, then `atom.value` type will be inferred from it and won't have `undefined` in its type.
~~2. If you specify generic type in `map` method as `Record` then `setKey` method won't allow you pass `undefined` as the second parameter and therefore delete item from the map store. It fixes this.~~

### DIFF
--- a/atom/index.d.ts
+++ b/atom/index.d.ts
@@ -81,7 +81,7 @@ export interface ReadableAtom<Value = any> {
    * Low-level method to read store’s value without calling `onStart`.
    *
    * Try to use only {@link ReadableAtom#get}.
-   * Without subscribers, value can de undefined.
+   * Without subscribers, value can be undefined.
    */
   readonly value: undefined | Value
 }
@@ -100,6 +100,11 @@ export interface WritableAtom<Value = any> extends ReadableAtom<Value> {
    * @param newValue New store value.
    */
   set(newValue: Value): void
+}
+
+export interface PreinitializedWritableAtom<Value extends any>
+  extends WritableAtom<Value> {
+  readonly value: Value
 }
 
 export type Atom<Value = any> = ReadableAtom<Value> | WritableAtom<Value>
@@ -136,4 +141,4 @@ export declare let notifyId: number
  */
 export function atom<Value, StoreExt = {}>(
   ...args: undefined extends Value ? [] | [Value] : [Value]
-): StoreExt & WritableAtom<Value>
+): PreinitializedWritableAtom<Value> & StoreExt
--- a/map/index.d.ts
+++ b/map/index.d.ts
@@ -117,6 +117,11 @@ export interface MapStore<Value extends object = any>
   ): () => void
 }
 
+export interface PreinitializedMapStore<Value extends object = any>
+  extends MapStore<Value> {
+  readonly value: Value
+}
+
 /**
  * Create map store. Map store is a store with key-value object
  * as a store value.
@@ -126,4 +131,4 @@ export interface MapStore<Value extends object = any>
  */
 export function map<Value extends object, StoreExt extends object = {}>(
   value?: Value
-): MapStore<Value> & StoreExt
+): PreinitializedMapStore<Value> & StoreExt
--- a/atom/errors.ts
+++ b/atom/errors.ts
@@ -1,4 +1,4 @@
-import { atom, WritableAtom } from '../index.js'
+import { atom, ReadableAtom } from '../index.js'
 
 let $store = atom<{ value: string }>({ value: '1' })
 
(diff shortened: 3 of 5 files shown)


map/index.d.ts
```diff
@@ -89,7 +89,7 @@ export interface MapStore<Value extends object = any>
    */
   setKey<Key extends AllKeys<Value>>(
     key: Key,
-    value: Get<Value, Key> | Value[Key]
+    value: Get<Value, Key> | undefined | Value[Key]
```
ai, line 92: This change will allow removing key which it is prohibited to remove:

```ts
let $store = map<{ required: string, optional?: string }>({ required: 'a' })
$store.setKey('optional', undefined) // OK
$store.setKey('required', undefined) // Expected to throw
```

You should not use just `Record` if you allow removing any key.

psd-coder: Yeah, you are right, it breaks types for literal objects...

> You should not use just Record if you allow removing any key.

What do you mean, can you elaborate? I have such code:
```
type Id = string;
type HeightMap = Record<Id, number>

const $heightMap = map<HeightMap>();
```

And it doesn't allow setting `undefined` to remove key from map.
```
// Argument of type `undefined` is not assignable to parameter of type `number`
$heightMap.setKey('id', undefined)
```

Should I use `map` here at all?

ai: I think will solve the problem `{ [key: string]?: number }`

psd-coder: This will break types in other places 🙂. For example, you have helper for getting total height from map:
```
function calcTotalHeight(heightMap: HeightMap) {
   return Object.values(heightMap).reduce((acc, height) => acc + height, 0);
}
```
With your suggestion height might be undefined, but in reality it can't. [Proof](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAkgJlAvFAzsATgSwHYHMDcAUKJFABISa4AWwAsgIZhJQBKEAxgPbpwA88ADRRsAVwC2AIwjoAfERLQKVWozAB5MMExdsDADYsACg3TaDfZTXpNZhQgDNR2Dtt1QOBjgBUuwA1a0ABTUlNZqAFzkYapMAJRQAN6EUKnoEMCi6NhQ6pIAVpzAAHQAbgaiECghMTZgccXpcKIcEEFBDBwcwqEqwAmIslCdHFAA1FC91sIADHFEAL72Ti5uOZ76Pn4BtZprBjV9kdFHTHs6evoJyalQ6ZnZuQVFZRVVh+HxjRDNre0jPVqAyGI3Gk1qs3mhAWQA)

It looks like the root of problem in the ambiguity of the `setKey` method:
1. It considers `undefined` as special value and doesn't set key, but removes it instead (which is contradicts to method name, btw)
2. It doesn't allow `undefined` as values at all.

Have you thought about separate method for removing key, like `removeKey/deleteKey`? It looks like it would solve all these issues

ai: What type you see for `removeKey`?

psd-coder: After some investigations I understood that it will have the same issue, because `Record` type is just syntax sugar for object literal, and it seems it isn't possible to differentiate them. 
Ideologically,  I think the current `.map` implementation is more likely `.object` or `.record`, and `.map` must work with `Record` type instead, which is basically map between key and value.
I don't think you want to change API in this way, so I can just drop that commit and leave the other one. Wdyt?

atom/index.d.ts
```diff
@@ -81,9 +81,9 @@ export interface ReadableAtom<Value = any> {
    * Low-level method to read store’s value without calling `onStart`.
    *
    * Try to use only {@link ReadableAtom#get}.
-   * Without subscribers, value can de undefined.
+   * Without subscribers, value can be undefined.
    */
-  readonly value: undefined | Value
+  readonly value: undefined extends Value ? undefined | Value : Value
```
ai, line 86: Will it work with `computed()`? I think `computed` will have `store.value === undefined` before the subscriber even if the origin type has no `undefined`

psd-coder: Yep, it won't because [`undefined` is explicitly set as initial value for `atom` in `computed`](https://github.com/nanostores/nanostores/blob/main/computed/index.js#L30). Looks like we can set `stores.map(store => store.value)` instead, but it will work only for sync callbacks.

ai: We can add some extra type like `PreinitializedAtom`

ai: BTW, you can use `$store.get()` to avoid `?`

psd-coder: Added `PreinitializedAtom` util as you suggested.

> BTW, you can use $store.get() to avoid ?

But it will cause subscription, isn't it?

ai: > But it will cause subscription, isn't it?

Only if there is no other listeners.

> Added PreinitializedAtom util as you suggested.

Did you push? Don't see in PR

psd-coder: Sorry :) Now it is pushed

map/index.d.ts
```diff
@@ -37,7 +38,7 @@ export type MapStoreKeys<SomeStore> = SomeStore extends {
   : AllKeys<StoreValue<SomeStore>>
 
 export interface MapStore<Value extends object = any>
-  extends WritableAtom<Value> {
+  extends PreinitializedWritableAtom<Value> {
```
ai, line 41: No, we can’t add it to `MapStore` because this type can be used by someone else.

Let’s add `PreinitializedMap`.

psd-coder: Added `PreinitializedMapStore`

Issue #326, opened 2024-08-05
Error since release `0.11.0` - The inferred type of 'XXX' cannot be named without a reference... 


Johannes-Andersen:
Hey!

Since the 0.11.0 all our stores are now failing in TS. Not sure if I am doing something wrong, but even coping the example from the readme is failing.

Error:
```
The inferred type of '$users' cannot be named without a reference to '../../../../node_modules/nanostores/atom'. This is likely not portable. A type annotation is necessary.ts(2742)
```

Example code:
```
interface User {
  id: string
  name: string
}

export const $users = atom<User[]>([])
```

![Screenshot 2024-08-05 at 11 07 00](https://github.com/user-attachments/assets/034dad65-bfcb-4182-87fc-a25f4c17b993)

ai:
cc @psd-coder could be relatated to your PR

gismya:
Reverting the changes in #322 fixes the issue.

I'll make a code sandbox that exposes the issues

ai:
I need a repo to reproduce the error.

Nothing happens in my project.

ai:
Also check your TypeScript version and `tsconfig.json`.

At least this file:

```ts
import { atom } from 'nanostores'

interface User {
  id: string
  name: string
}

export const $users = atom<User[]>([])
```

with this config:

```json
{
  "compilerOptions": {
    "target": "es2018",
    "module": "esnext",
    "moduleResolution": "node",
    "esModuleInterop": true,
    "allowJs": true,
    "strict": true,
    "noEmit": true,
    "jsx": "react"
  }
}
```

Have no errors:

```
tsc --noEmit
```

gismya:
I'm nailing down the issues. It's depending on the TSconfig, and I'm currently trying to find which option it is. I think it's the `moduleResolution`

gismya:
Okay, testing different options in the TSconfig it seems like your example TSconfig, but with changes

```
    "moduleResolution": "NodeNext",
    "declaration": true,
```

is what's needed to trigger the issue.

https://codesandbox.io/p/sandbox/nanostores-typescript-issue-sx4kgm?file=%2Fsrc%2Fnanostores.ts

ai:
Fixed by https://github.com/nanostores/nanostores/commit/1c5636df6829f9087f8a9a895b48a71cdf440104 and https://github.com/nanostores/nanostores/pull/327

The fix was released in 0.11.2

PR #327, opened 2024-08-05
fix(TS): inferred type cannot be named without reference

gismya:
### Fixes Issue #326

This PR addresses an issue where type inference fails under the following TSconfig settings:

```json
{
    "moduleResolution": "NodeNext",
    "declaration": true
}
```

Since the implementation of #322, the new `Preinitialized` types were not being exported from the module root, causing the type inference to break.

A CodeSandbox example illustrating the issue can be found [here](https://codesandbox.io/p/sandbox/nanostores-typescript-issue-sx4kgm?file=%2Fsrc%2Fnanostores.ts).

**Changes in this PR:**

- Adds the missing exports for Preinitialized types to ensure proper type inference.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,10 @@
-export { atom, Atom, ReadableAtom, WritableAtom } from './atom/index.js'
+export {
+  atom,
+  Atom,
+  PreinitializedWritableAtom,
+  ReadableAtom,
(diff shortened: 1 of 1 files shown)


ai:
Thanks. I already fixed `atom` but forgot about `map`.

PR #330, opened 2024-08-20
Add `effect` method

psd-coder:
# Motivation

In spite of having ability to use atom's subscribe/notify methods for subscribing on changes, it might be a bit tricky to track changes at multiple atoms at once. `effect` is designed to solve this problem! And it is similar to the `effect` method from other state managers (like Solid's signals, or state manager in Svelte)

# How it works

`effect` runs its callback on the start, with initial values, as well as on any store(s) change. If callback returns cleanup function it will be performed before next callback run. Besides that, `effect` returns own cleanup function, which allows cancelling the whole effect.

# Usages

`effect` can be used as analogue of `subscribe` method on one store:

```ts
import { atom, effect } from 'nanostores'

const $num = atom(10)

effect($atom, num => {
  console.log(`Number is ${num}`)
})
```

`effect` allows easily handle complex cases. This example shows how we can create resize observer atom, which will create [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) instance based on passed element and options (which can be changes outside, since they are both atoms). Whenever element or options are changed it re-runs effect and cleans up ResizeObserver instance for the previous effect run.

```ts
import { atom, effect, onMount, type ReadableAtom } from 'nanostores'

type ResizeObserverAtomOptions<E extends HTMLElement, Mapped> = {
  $element: ReadableAtom<E | null>
  $observerOptions?: ReadableAtom<ResizeObserverOptions>
  initialValue: Mapped
  mapper: (entry: ResizeObserverEntry) => Mapped
}

export function createResizeObserverAtom<E extends HTMLElement, Mapped>({
  $element,
  $observerOptions,
  initialValue,
  mapper
}: ResizeObserverAtomOptions<E, Mapped>) {
  const $result = atom(initialValue)

  onMount($result, () => {
    // Notice how we use returned cleanup function as returned function from onMount
    // It will automatically stop effect when $result no longer has any subscribers,
    // and cleanup everything
    return effect([$element, $observerOptions], (element, observerOptions) => {
      if (!element) {
        return
      }

      const resizeObserver = new ResizeObserver(entries => {
        const entry = entries[0]

        if (entry) {
          $result.set(mapper(entry))
        }
      })
      resizeObserver.observe(element, observerOptions)

      return () => resizeObserver.disconnect()
    })
  })

  return $result
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ export const Admins = () => {
   - [Deep Maps](#deep-maps)
   - [Lazy Stores](#lazy-stores)
   - [Computed Stores](#computed-stores)
+  - [Effects](#effects)
   - [Tasks](#tasks)
   - [Store Events](#store-events)
 - [Integration](#integration)
@@ -390,6 +391,42 @@ import { $posts } from './posts.js'
 
 export const $newPosts = computed([$lastVisit, $posts], (lastVisit, posts) => {
   return posts.filter(post => post.publishedAt > lastVisit)
+})
+```
+
+
+### Effects
+
+In spite of having ability to use atom's subscribe/notify methods for subscribing on changes, it might be a bit tricky to track changes at multiple atoms at once. `effect` is designed to solve this problem!
+
+`effect` runs its callback on the start, with initial values, as well as on any store(s) change. If callback returns cleanup function it will be performed before next callback run. Besides that, `effect` returns own cleanup function, which allows cancelling the whole effect.
+
+`effect` can be used as analogue of `subscribe` method on one store:
+
+```ts
+import { atom, effect } from 'nanostores'
+
+const $num = atom(10)
+
+effect($atom, num => {
+  console.log(`Number is ${num}`)
+})
+```
+
+`effect` allows easily handle complex cases. This example shows how you can execute any method with specific interval and only when it is enabled. Both `$isEnabled` and `$interval` are atoms and their values can be changed. Whenever it happens it clears interval from the previous run.
+
+```js
+import { sendPing } from '~/api'
+
+const $isEnabled = atom(true)
+const $interval = atom(1000)
+
+const cancelPing = effect([$isEnabled, $interval], (isEnabled, interval) => {
+  if (!isEnabled) return
+
+  const intervalId = setInterval(() => sendPing(), interval)
+
+  return () => clearInterval(intervalId)
 })
 ```
 
--- a/effect/index.d.ts
+++ b/effect/index.d.ts
@@ -1,0 +1,33 @@
+import type { StoreValues } from '../computed/index.d.ts'
+import type { AnyStore, Store, StoreValue } from '../index.js'
+
+interface Effect {
+  <OriginStore extends Store>(
+    stores: OriginStore,
+    cb: (value: StoreValue<OriginStore>) => void | VoidFunction
+  ): VoidFunction
+  /**
+   *
+   * Create effect, which subscribes to all stores and runs callback with their values on change in any of them.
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -393,6 +382,69 @@ export const $newPosts = computed([$lastVisit, $posts], (lastVisit, posts) => {
...
+`effect` can be used as analogue of `subscribe` method on one store:
+
+```ts
+import { atom, effect } from 'nanostores'
+
+const $num = atom(10)
+
+effect($atom, num => {
+  console.log(`Number is ${num}`)
+})
```
ai, line 400: We will need multi storage subscriber example (and link from the first mention of `listen` to answer the question how to listen multiple stores).

README.md
```diff
@@ -393,6 +382,69 @@ export const $newPosts = computed([$lastVisit, $posts], (lastVisit, posts) => {
...
+```
+
+`effect` allows easily handle complex cases. This example shows how we can create resize observer atom, which will create [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) instance based on passed element and options (which can be changes outside, since they are both atoms). Whenever element or options are changed it re-runs effect and cleans up ResizeObserver instance for the previous effect run.
+
+```ts
```
ai, line 405: Too complex example for the docs

effect/index.d.ts
```diff
@@ -0,0 +1,40 @@
...
+    cb: (value: StoreValue<OriginStore>) => void | VoidFunction
+  ): VoidFunction
+  /**
+   *
+   * Create effect, which subscribes to all stores and calls callback with their values on change in any of them.
+   * Especially useful for side effects which must cleanup after themselves.
+   *
+   * Example of resize observer effect, which depends on the element and observer options atoms
+   * ```js
+   * let $height = atom(0)
```
ai, line 16: It will be impossible to see this example in IDE popup

effect/index.test.ts
```diff
@@ -0,0 +1,79 @@
+import { strictEqual } from 'node:assert'
+import type { TestContext } from 'node:test'
+import { test } from 'node:test'
+
+import { atom } from '../atom/index.js'
+import { effect } from './index.js'
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
ai, line 8: Why not to define return type? It speeds up TS check.

effect/index.d.ts
psd-coder: Why is it? I see it normarly. Or do you mean it is too big as an example in JSDoc?

README.md
psd-coder: I mentioned `listen` as a way to subscribe on changes, not as technically the same method (since it doesn't run callback on initialization).
If you think it might confuse users, let's just remove this mention, because otherwise example will require separate counter/boolean flag to exclude initial run. Like:

```js
const $atom1 = atom(1);
const $atom2 = atom(2);

let isFirstRun = true;
effect([$atom1, $atom2], (atom1, atom2) => {
  if (isFirstRun) {
    isFirstRun = false
    return
  }

  console.log(atom1 + atom2);
});
```

We could add third parameter, to the `effect` to have control over this, but I believe it is rare case, and it can be easily handled in above-mentioned way. I prefer to not overcomplicate API (I don't remember such option in other state managers either, btw)

effect/index.d.ts
psd-coder: Simplified examples, both in README and JSDoc. Could you check it?

ai:
Thanks. Great idea.

PR #333, opened 2024-08-26
Fix types for computed with async task

ai:
The bug found by @psd-coder


ai:
It was released in 0.11.3.

PR #336, opened 2024-09-15
Explicitly allow undefined as setKey input

seangenabe:
In TypeScript, allows `undefined` as input to `map.setKey` as documented.

>    To delete key set `undefined`.
>    
>    ```js
>    $settings.setKey('theme', undefined)
>    ```

### DIFF
--- a/map/index.d.ts
+++ b/map/index.d.ts
@@ -97,7 +97,7 @@ export interface MapStore<Value extends object = any>
    */
   setKey<Key extends AllKeys<Value>>(
     key: Key,
-    value: Get<Value, Key> | Value[Key]
+    value: Get<Value, Key> | Value[Key] | undefined
   ): void
 
   /**


ai:
This change will allow to delete mandatory keys

```
let $user = map<{ login: string, name?: string }>({ login: 'foo' })

$user.setKey('name', undefined) // OK, optional key
$user.setKey('login', undefined) // should throw an error since $user should always have login
```

seangenabe:
Without this change, TypeScript users won't be able to pass `undefined` to this function in the first place. Is there an alternative solution for this?

ai:
@seangenabe if you define your key as optional `login?: string` if will be able to delete it.

Show your store source code if you have any problem.

seangenabe:
You're absolutely right. I think I was just looking at this the wrong way. My bad.

What I have is a generic map of strings to objects, e.g. `Record<string, Something>>`. I guess I could just make it `Partial`.

Thank you and I apologize this has kinda turned into a support request.

PR #356, opened 2025-02-28
Add getKey() function

gismya:


ai:
Hm. I just found that we already have `getPath()`. Maybe we can fix it a little to accept `Map`?

PR #357, opened 2025-03-02
Allow undefined for map and deepmap

gismya:
Allows undefined values for `map()` and `deepMap()`.

Has a known type issue

### DIFF
--- a/deep-map/index.d.ts
+++ b/deep-map/index.d.ts
@@ -1,9 +1,16 @@
 import type { WritableAtom } from '../atom/index.js'
 import type { AllPaths, BaseDeepMap, FromPath } from './path.js'
 
-export { AllPaths, BaseDeepMap, FromPath, getPath, setByKey, setPath } from './path.js'
+export {
+  AllPaths,
+  BaseDeepMap,
+  FromPath,
+  getPath,
+  setByKey,
+  setPath
+} from './path.js'
 
-export type DeepMapStore<T extends BaseDeepMap> = {
+export type DeepMapStore<T extends BaseDeepMap | undefined> = {
   /**
    * Subscribe to store changes.
    *
@@ -78,4 +85,6 @@ export type DeepMapStore<T extends BaseDeepMap> = {
  * @param init Initialize store and return store destructor.
  * @returns The store object with methods to subscribe.
  */
-export function deepMap<T extends BaseDeepMap>(init?: T): DeepMapStore<T>
+export function deepMap<Value extends BaseDeepMap | undefined>(
+  value?: Value
+): DeepMapStore<Value>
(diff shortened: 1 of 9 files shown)


map/errors.ts
```diff
@@ -38,4 +38,26 @@ $test.setKey('b', 5)
...
+// Subscribe picks up the undefined type of the store
+$test2.subscribe(value => {
+  // THROWS 'value' is possibly 'undefined'
+  value.isLoading
+})
+
+// THROWS Argument of type undefined is not assignable to parameter of type 'Record<string, unknown>'.
+let $test3 = map<Record<string, unknown>>(undefined) // Currently doesn't throw
+// THROWS Expected 1 arguments, but got 0
+let $test4 = map<Record<string, unknown>>() // Currently doesn't throw
```
gismya, line 60: The known issues is that these do not throw

deep-map/index.d.ts
```diff
@@ -78,4 +85,6 @@ export type DeepMapStore<T extends BaseDeepMap> = {
  * @param init Initialize store and return store destructor.
  * @returns The store object with methods to subscribe.
  */
-export function deepMap<T extends BaseDeepMap>(init?: T): DeepMapStore<T>
+export function deepMap<Value extends BaseDeepMap | undefined>(
+  value?: Value
```
ai, line 89: We need to allow miss `value` only if `Value` type allows `undefined`

gismya: Yes, that's the part I couldn't figure out. Not sure if it's the function overloading syntax I'm not getting right or if it requires some other sort of type wrangling.

ai: We can use two definitions like:

```ts
export function map<Value extends object>(value: Value): PreinitializedMapStore<Value>
export function map<Value extends undefined>(value?: Value): PreinitializedMapStore<Value>
```

gismya: I feel that was exactly what I tried, but I'll have another look.

ai: Try to change the order, maybe we need to put `undefined` first

gismya: No, doesn't matter. Neither of them works. Also tried having then, and a third one that covers both cases. 

Not having one that covers both makes both versions complain:

`Type '{ a: number; }' does not satisfy the constraint 'undefined'.`
`Type 'Record<string, unknown> | undefined' does not satisfy the constraint 'object'.`

And adding the overload that covers all three then swallows the cases that should be errors. Not sure if it's a restriction in having union overloads with the same number of parameters (Similar to https://github.com/microsoft/TypeScript/issues/14107)

gismya:
I'm closing this for now. Not a feature I need, and I don't know if we can fix this without remaking the API. Anyone else who is motivated is free to pick this up instead.

PR #358, opened 2025-03-02
Allow setKey undefined on string indexed maps

gismya:
The most asked question I've seen for this package is about the behavior of `setKey` on maps, when it comes to `undefined`. For string-indexed maps that don't explicitly allow undefined, users need to use type assertions to delete items. (For example, check out [this question](https://github.com/orgs/nanostores/discussions/353#discussioncomment-12338911)).

This behavior doesn’t really match how TypeScript handles string-indexed maps. While I think Nanostores current approach is technically more correct than TypeScript, it’s clearly not the behavior users expect (And it would break too many repos for TS to change it's default behavior, see [this discussion](https://github.com/microsoft/TypeScript/issues/49169)).

Therefore I came up with a solution where we can check if the Map is string indexed, and in that case allow deletion using undefined. 

Right now, this PR is just a discussion. If we decide this is the direction to go, I’ll also need to add a similar change to the `deepMap`.

### DIFF
--- a/map/errors.ts
+++ b/map/errors.ts
@@ -37,5 +37,9 @@ $test.setKey('b', 5)
 // THROWS Argument of type '"z"' is not assignable to parameter
 $test.setKey('z', '123')
 
+let $testIndexSignature = map<Record<string, number>>()
+$testIndexSignature.setKey('a', 1)
+$testIndexSignature.setKey('a', undefined)
+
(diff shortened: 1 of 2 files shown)


ai:
Thanks. Yes, for the case of `Record` we should allow deleting keys since there is no clear must-have keys.

PR #359, opened 2025-03-03
Allow set key undefined on string indexed deepMap

gismya:
Following up on #358, this PR adds typings for `deepMap`. I haven't used `deepMap`, so there might be some edge cases I've missed, but all of the changes are to the types.

The typings for the deep pathing is complicated, and it essentially forced me to duplicate the types that were affected. I also tried adding thorough type tests (Root level indexed, lower level indexed, making sure the base case was not affected), but if there are any cases that I didn't test I'd gladly add that.


ai:
Thanks! Amazing work

PR #373, opened 2025-11-07
 Deepmap: Modify setKey for support merge items and save previous value

Lifo123:
### Make `setKey` smarter with an `options`  params

The main purpose of adding an `options` parameter to `setKey` is to **improve the developer experience (DevEx)** by making the function more versatile.

-----

#### The Problem (Before)

Right now, `setKey` is a simple **setter**. It just *replaces* a value.

If you want to *update* a nested object like `target` you have to write boilerplate code in your component:

```ts
// Boilerplate: Read, Merge, then Set
function updateTarget(props) {
  const { target } = $mapStore.get(); // 1. Read
  const updatedTarget = { ...target, ...props }; // 2. Merge manually
  $mapStore.setKey("target", updatedTarget); // 3. Set
}
```

This is imperative—you are telling the store *how* to do its job.

#### The Solution (After)

With an `options` prop, `setKey` becomes a smart **updater**.

```ts
// Declarative: Just tell the store to merge
function updateTarget(props) {
  $mapStore.setKey("target", props, { shouldMerge: true });
}
```

This is declarative, you just tell the store *what* you want and make cleaning for read code in frameworks.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,7 +36,7 @@ jobs:
           echo "EOF" >> $GITHUB_OUTPUT
       - name: Create the release
         if: steps.changelog.outputs.changelog_content != ''
-        uses: softprops/action-gh-release@72f2c25fcb47643c292f7107632f7a47c1df5cd8 # v2.3.2
+        uses: softprops/action-gh-release@6da8fa9354ddfdc4aeace5fc48d7f679b5214090 # v2.4.1
         with:
           name: ${{ github.ref_name }}
           body: '${{ steps.changelog.outputs.changelog_content }}'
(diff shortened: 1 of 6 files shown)


Lifo123:
Yes, of course, I really like using DeepMap and I agree with keeping the core simple.

So @nanostores/deep-map could be a good option to keep it within the ecosystem.

ai:
The plan:

1. I sent you invite
2. You take `nanostores` repo, delete all files expect `deepMap` (to keep it history)
3. And push to `nanostores/deep-map`
4. I review the repo and will give you npm persmission when it will be ready

Then we will deprecate `deepMap` in 1.1 and start to promote `nanostores/deep-map`. But will delete old deprecated `deepMap` only in 2.0.

gismya:
I fully support the idea of moving deepMap from the core repo. I was a bit uncomfortable with how different if felt to the rest of the package in terms of complexity. Good decision and thanks @Lifo123 for wanting to maintain it.

PR #396, opened 2026-04-15
Fix: store epoch on globalThis to prevent separate module instances from diverging

### LINKED ISSUE #371: Issue with "epoch" Global Variable Across Multiple Module Instances
In `nanostores/atom`, there is a global variable called `epoch` that is used in `nanostores/computed` to determine whether state changes should be computed.

However, if two dependent modules both use nanostores for reactive state, they could each end up with their own instance of the `epoch` variable. This can break application logic, as the state change detection relies on the shared value of `epoch`.

**Expected behavior:**  
There should be a single, consistent instance of `epoch` across all modules using nanostores, to ensure correct computation of state changes.

**Actual behavior:**  
Multiple instances of `epoch` may be created if nanostores is bundled separately in different parts of an application, leading to inconsistent state updates.

To fix this, maybe store `epoch` on `globalThis` using a symbol, like:

```js
const EPOCH_SYMBOL = Symbol.for("nanostores.epoch")
globalThis[EPOCH_SYMBOL] = globalThis[EPOCH_SYMBOL] || 0
```

### Extra context
In our setup, we've got a monorepo with multiple libraries—a core (vanilla JS) version and a React version. The React version uses `computed`, but the core version doesn't.

Because our bundler has tree shaking enabled, `epoch` was getting removed from the core bundle since nothing in core actually reads its value. The React version uses `computed` to subscribe to multiple stores from core and compute values from them.

Here's where things break: since the core and React versions are bundled separately, only the React bundle has the `epoch` logic. The stores themselves come from core (which is missing `epoch`), so the `computed` in React doesn't get updates from those stores. React ends up with its own `epoch` instance, which isn't being updated by changes in core, causing computed values to never update.

afurm:
## What

Stores the shared `epoch` counter on `globalThis.nanostores` instead of as a plain module-level variable. This ensures all module instances of nanostores share the same epoch counter, regardless of how they are bundled.

## Why

When nanostores is bundled separately in different parts of an application (e.g., tree-shaking separates core from React, or Turbopack creates multiple module instances — see https://github.com/nanostores/nanostores/issues/371#issuecomment-2098405189), each bundle can end up with its own `epoch` instance. This causes `computed` stores to not receive updates from `atom` stores in different bundles, because the epoch counters are independent and never sync.

## How

The fix stores `epoch` on `globalThis.nanostores` (a shared namespace object) before exporting it:

This approach:
1. Uses `globalThis` as a truly global storage (not affected by bundler tree-shaking)
2. Creates a shared `nanostores` namespace on `globalThis` to avoid polluting the global namespace
3. Initializes `epoch` only if not already present (safe for multiple module instances)
4. Maintains the same `export let epoch` API so no changes needed elsewhere

## Testing

All 46 existing tests pass with this change.

Closes #371


atom/index.js
```diff
@@ -3,7 +3,14 @@ import { clean } from '../clean-stores/index.js'
 let listenerQueue = []
 let lqIndex = 0
 const QUEUE_ITEMS_PER_LISTENER = 4
-export let epoch = 0
+// Use globalThis.nanostores to store epoch so all module instances share the same counter.
+// This fixes issues when nanostores is bundled separately in different parts of an
+// application (e.g., tree-shaking separates core from React), causing each bundle
+// to have its own epoch instance.
+// See: https://github.com/nanostores/nanostores/issues/371
+let NANOSTORES = globalThis.nanostores || (globalThis.nanostores = {})
```
ai: We should use at least `nanostoresEpoch`

atom/index.js
```diff
@@ -3,7 +3,13 @@ import { clean } from '../clean-stores/index.js'
 let listenerQueue = []
 let lqIndex = 0
 const QUEUE_ITEMS_PER_LISTENER = 4
-export let epoch = 0
+// Use globalThis.nanostores to store epoch so all module instances share the same counter.
+// This fixes issues when nanostores is bundled separately in different parts of an
+// application (e.g., tree-shaking separates core from React), causing each bundle
+// to have its own epoch instance.
+// See: https://github.com/nanostores/nanostores/issues/371
+let NANOSTORES = globalThis.nanostoresEpoch || (globalThis.nanostoresEpoch = { epoch: 0 })
```
ai: We need to update `package.json`→`size-limit` to see the price

atom/index.js
```diff
@@ -3,7 +3,13 @@ import { clean } from '../clean-stores/index.js'
 let listenerQueue = []
 let lqIndex = 0
 const QUEUE_ITEMS_PER_LISTENER = 4
-export let epoch = 0
+// Use globalThis.nanostores to store epoch so all module instances share the same counter.
```
ai: We need to update comment

atom/index.js
afurm: Acknowledged. The code already uses nanostoresEpoch.

atom/index.js
afurm: Fixed. Updated Atom limit to 311 B and Popular Set limit to 835 B in package.json.

atom/index.js
afurm: Fixed. Comment now correctly references globalThis.nanostoresEpoch.

gismya:
Correct me if I'm wrong, but I don't think this works? `export let epoch = NANOSTORES.epoch` should only copy the current value of the `NANOSTORES.epoch` since the `epoch` itself is a primitive.

atom/index.js
```diff
@@ -3,7 +3,13 @@ import { clean } from '../clean-stores/index.js'
...
 let lqIndex = 0
 const QUEUE_ITEMS_PER_LISTENER = 4
-export let epoch = 0
+// Use globalThis.nanostoresEpoch to store epoch so all module instances share
+// the same counter. This fixes issues when nanostores is bundled separately
+// in different parts of an application (e.g., tree-shaking separates core
+// from React), causing each bundle to have its own epoch instance.
+// See: https://github.com/nanostores/nanostores/issues/371
+let NANOSTORES = globalThis.nanostoresEpoch || (globalThis.nanostoresEpoch = { epoch: 0 })
+export let epoch = NANOSTORES.epoch
```
gismya:
```suggestion
export let epoch = globalThis.nanostoresEpoch ||= { epoch: 0 }
```
And changing all instances in the code of `epoch` to `epoch.epoch` should solve the primitive not being shared. Saving it at the same time as instantiating with logical OR assignment brings the size down.

gismya: Testing that gives a size of `296 B` for the Atom and `830 B` for the popular set.

gismya:
We should also add a test for the multiple epoch scenario

atom/index.js
ai: I am also not sure that `let epoch = NANOSTORES.epoch` works.

But we can also:

```js
globalThis.__NANOSTORES_GLOBAL ||= { epoch: 0 }
export const NANOSTORES_GLOBAL = globalThis.__NANOSTORES_GLOBAL
```

export object here. `epoch` is not a public API if I remember it correctly.

afurm: Fixed per your suggestion. The bug was that `export let epoch = NANOSTORES.epoch` captures the primitive value at init time, so `epoch++` increments the local copy, not the shared global.

Changes:
- Export `NANOSTORES_EPOCH` (the shared object) instead of the primitive epoch value
- Use `NANOSTORES.epoch++` in notify() to increment the shared counter
- Updated `computed/index.js` to use `NANOSTORES_EPOCH.epoch`

All 46 tests pass. Size limits updated: Atom 307 B (under 311 B limit), Popular Set 841 B (under 845 B limit).

afurm: Thanks for the suggestion! Applied the ||= optimization which saves 11 B on Atom (now 300 B) and 8 B on Popular Set (now 837 B). All tests pass.

afurm: Your measured sizes were very close. With the ||= optimization and our NANOSTORES_EPOCH approach, we get Atom 300 B and Popular Set 837 B — all within limits.

ai:
Thank. The fix was released in 1.3.0.

PR #397, opened 2026-04-15
Use nanostoresEpoch as global key

afurm:
Fixes https://github.com/nanostores/nanostores/pull/396#discussion_r3087863707


ai:
It is better to push commit to origin PR.

PR #399, opened 2026-05-21
Add `batch` primitive for grouping writes

psd-coder:
## Overview
 
Adds `batch(fn)` — a synchronous primitive that groups multiple store writes so each listener fires at most once when the outermost batch exits, observing the final state.
 
## Problem Statement
 
Nanostores has no way to group multiple synchronous writes. When several stores (or the same store) are updated in sequence, every listener and downstream computed runs after each individual `set()`. This produces redundant work, intermediate values being observed by effects, and forces consumers to choose between awkward "update one big object" patterns or wearing the cost of N cascades for N writes.
 
Without a batching primitive, library authors also can't safely batch internally — wrapping a helper's writes in any deferred-flush mechanism prevents callers from composing their own batches around it.
 
## Solution Approach
 
- Adds `batch(fn)` exported from the root. While inside a batch, `atom.notify` defers the listener-queue drain. On the outermost exit, the queue drains with per-listener deduplication so each listener fires once with the final atom value.
- Listener tuples now carry the atom reference rather than a snapshot of its value at push time, so the deduped firing reads the current value via `$atom.value` at drain time. `oldValue` is preserved from the first push, giving the natural "pre-batch → post-batch" transition.
- Inside a batch, dedup applies to every listener: each fires once with `changedKey === undefined`, so repeated or multi-key `map.setKey` writes coalesce like a plain `set`. This reuses the existing `map.set(newObject)` case, where `changedKey` is already undefined, so `listenKeys` subscribers behave exactly as they do for a full `map.set`. Behavior outside a batch is unchanged. Passing an array of changed keys as the third listener argument is left for the next major.
- Nesting is supported and exception-safe: inner batches are no-ops for flush purposes; the outermost `try/finally` always drains and clears the dedup set, including when `fn` throws. Note this also means `batch` is not a transaction: writes are not rolled back if `fn` throws, only their notifications are grouped.
- README gains a "Batching" section and TOC entry. `atom/index.d.ts` declares the new export, and root `index.{js,d.ts}` re-export it.
## Breaking Changes
 
None. Behavior outside a `batch(...)` call is unchanged — every existing test passes, including the diamond suite.
 
## Testing
 
- New tests in `atom/index.test.ts` cover the core semantics — coalescing, nesting, exception safety, computed and effect under batch, `map.setKey` + `listenKeys` correctness — and `computed/index.test.ts` adds batched-diamond cases.
## Size change
 
+46 B for Atom / +33 B for Popular set.

### DIFF
--- a/atom/index.d.ts
+++ b/atom/index.d.ts
@@ -172,3 +172,30 @@ export function atom<Value, StoreExt = object>(
 export function readonlyType<Value>(
   store: ReadableAtom<Value>
 ): ReadableAtom<Value>
+
+/**
+ * Run multiple store changes as a single transaction. While inside `batch`,
+ * listeners are deferred and each listener fires at most once when the
+ * outermost `batch` returns — with the final store values.
+ *
+ * Batches can be nested; only the outermost batch triggers the flush.
+ *
+ * Synchronous only. The batch closes when `fn` returns, so writes performed
+ * in microtasks, timers, or after `await` will not be batched. Pass a sync
+ * function.
+ *
+ * ```ts
+ * import { atom, batch } from 'nanostores'
+ *
+ * let $a = atom(0)
+ * let $b = atom(0)
+ *
+ * batch(() => {
+ *   $a.set(1)
+ *   $b.set(2)
+ * })
+ * ```
+ *
+ * @param fn Callback that performs the store writes.
+ */
+export function batch(fn: () => void): void
--- a/atom/index.d.ts
+++ b/atom/index.d.ts
@@ -184,6 +184,9 @@ export function readonlyType<Value>(
  * in microtasks, timers, or after `await` will not be batched. Pass a sync
  * function.
  *
+ * `Map#setKey` writes are coalesced as well, so the listener's `changed`
+ * argument is `undefined` for the batched notification.
+ *
  * ```ts
  * import { atom, batch } from 'nanostores'
  *
--- a/atom/index.js
+++ b/atom/index.js
@@ -2,12 +2,45 @@ import { clean } from '../clean-stores/index.js'
 
 let listenerQueue = []
 let lqIndex = 0
+let batchSeen = null
 const QUEUE_ITEMS_PER_LISTENER = 4
 // Use globalThis.nanostoresGlobal to store epoch so all module instances share
 // the same counter. This fixes issues when Nano Store is bundled separately
 // in different parts of an application (e.g., tree-shaking separates core
 // from React), causing each bundle to have its own epoch instance.
 export const nanostoresGlobal = (globalThis.nanostoresGlobal ||= { epoch: 0 })
+
+let drainQueue = () => {
+  for (
+    lqIndex = 0;
+    lqIndex < listenerQueue.length;
+    lqIndex += QUEUE_ITEMS_PER_LISTENER
+  ) {
+    listenerQueue[lqIndex](
+      listenerQueue[lqIndex + 1].value,
+      listenerQueue[lqIndex + 2],
+      listenerQueue[lqIndex + 3]
+    )
+  }
+  listenerQueue.length = 0
+}
+
+/* @__NO_SIDE_EFFECTS__ */
+export const batch = fn => {
+  let outer = !batchSeen
+  if (outer) batchSeen = new Set()
+  try {
+    fn()
+  } finally {
+    if (outer) {
+      try {
+        if (listenerQueue.length) drainQueue()
+      } finally {
+        batchSeen = null
+      }
+    }
+  }
+}
 
 /* @__NO_SIDE_EFFECTS__ */
 export const atom = initialValue => {
@@ -45,24 +78,15 @@ export const atom = initialValue => {
     },
     notify(oldValue, changedKey) {
       nanostoresGlobal.epoch++
-      let runListenerQueue = !listenerQueue.length
+      let runListenerQueue = !listenerQueue.length && !batchSeen
       for (let listener of listeners) {
-        listenerQueue.push(listener, $atom.value, oldValue, changedKey)
+        if (!changedKey && batchSeen?.has(listener)) continue
+        if (!changedKey) batchSeen?.add(listener)
+        listenerQueue.push(listener, $atom, oldValue, changedKey)
       }
 
       if (runListenerQueue) {
-        for (
-          lqIndex = 0;
-          lqIndex < listenerQueue.length;
-          lqIndex += QUEUE_ITEMS_PER_LISTENER
-        ) {
-          listenerQueue[lqIndex](
-            listenerQueue[lqIndex + 1],
-            listenerQueue[lqIndex + 2],
-            listenerQueue[lqIndex + 3]
-          )
-        }
-        listenerQueue.length = 0
+        drainQueue()
       }
     },
     /* It will be called on last listener unsubscribing.
--- a/atom/index.js
+++ b/atom/index.js
@@ -80,9 +80,14 @@ export const atom = initialValue => {
       nanostoresGlobal.epoch++
       let runListenerQueue = !listenerQueue.length && !batchSeen
       for (let listener of listeners) {
-        if (!changedKey && batchSeen?.has(listener)) continue
-        if (!changedKey) batchSeen?.add(listener)
-        listenerQueue.push(listener, $atom, oldValue, changedKey)
+        if (batchSeen?.has(listener)) continue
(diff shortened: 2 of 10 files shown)


atom/index.js
```diff
@@ -2,13 +2,46 @@ import { clean } from '../clean-stores/index.js'
...
+    listenerQueue[lqIndex](
+      listenerQueue[lqIndex + 1].value,
+      listenerQueue[lqIndex + 2],
+      listenerQueue[lqIndex + 3]
+    )
+  }
+  listenerQueue.length = 0
+}
+
+/* @__NO_SIDE_EFFECTS__ */
```
AleksandrSl, line 28: How ironic, it has a side effect writing to the module scope global, but I guess that's not what the bundler should be concerned about?

atom/index.js
```diff
@@ -45,24 +78,15 @@ export const atom = initialValue => {
     },
     notify(oldValue, changedKey) {
       nanostoresGlobal.epoch++
-      let runListenerQueue = !listenerQueue.length
+      let runListenerQueue = !listenerQueue.length && !batchSeen
       for (let listener of listeners) {
-        listenerQueue.push(listener, $atom.value, oldValue, changedKey)
+        if (!changedKey && batchSeen?.has(listener)) continue
+        if (!changedKey) batchSeen?.add(listener)
+        listenerQueue.push(listener, $atom, oldValue, changedKey)
```
AleksandrSl: So for two different changed keys, we will get two events anyway?

atom/index.d.ts
```diff
@@ -172,3 +172,26 @@ export function atom<Value, StoreExt = object>(
...
+ * let $a = atom(0)
+ * let $b = atom(0)
+ *
+ * batch(() => {
+ *   $a.set(1)
+ *   $b.set(2)
+ * })
+ * ```
+ *
+ * @param fn Callback that performs the store writes.
```
AleksandrSl, line 202: So the price of the implementations is temporarily inconsistent state, meaning you can get .value and will be already updated, but the listeners won't be notified

atom/index.test.ts
```diff
@@ -442,3 +451,200 @@ test('has readonly helper', () => {
...
+  })
+
+  deepStrictEqual(nameKey, ['name'])
+  deepStrictEqual(ageKey, ['age'])
+
+  unbindName()
+  unbindAge()
+})
+
+test('batch dedupes plain map.set but preserves setKey notifications', () => {
```
AleksandrSl: Not sure I get what the test name means, how do we check that setKey notifications are preserved here?

atom/index.test.ts
```diff
@@ -442,3 +451,200 @@ test('has readonly helper', () => {
...
+
+test('listenKeys still receives each changed key inside a batch', () => {
+  let $user = map<{ age?: number; name?: string }>({ age: 0, name: 'A' })
+  let nameKey: unknown[] = []
+  let ageKey: unknown[] = []
+  let unbindName = listenKeys($user, ['name'], (_v, _o, k) => nameKey.push(k))
+  let unbindAge = listenKeys($user, ['age'], (_v, _o, k) => ageKey.push(k))
+
+  batch(() => {
+    $user.setKey('name', 'B')
```
AleksandrSl, line 657: I'd also added some test cases where we update the same key multiple times. Could probably happen in the nested transactions?

atom/index.js
```diff
@@ -2,13 +2,46 @@ import { clean } from '../clean-stores/index.js'
...
+      listenerQueue[lqIndex + 3]
+    )
+  }
+  listenerQueue.length = 0
+}
+
+/* @__NO_SIDE_EFFECTS__ */
+export const batch = fn => {
+  let outer = !batchSeen
+  if (outer) batchSeen = new Set()
```
AleksandrSl, line 31: Could probably be faster if we check Set size and do not null it, but cleaning, but I'm not sure we care that much

atom/index.d.ts
```diff
@@ -172,3 +172,26 @@ export function atom<Value, StoreExt = object>(
...
+ * Batches can be nested; only the outermost batch triggers the flush.
+ *
+ * ```ts
+ * import { atom, batch } from 'nanostores'
+ *
+ * let $a = atom(0)
+ * let $b = atom(0)
+ *
+ * batch(() => {
+ *   $a.set(1)
```
AleksandrSl, line 197: What happens if we start abusing promises inside? 😂

AleksandrSl (commented):
Looks reasonable, not without the tradeoffs

atom/index.js
psd-coder: True :) It's a hint for the bundler's tree-shaking, so it should be safe to use it here.

atom/index.js
psd-coder: Yeah, It was done for backward compatibility. I don't like it much too, so let's discuss it with @ai

atom/index.d.ts
psd-coder: Yeah, you are right. We could store value from `set` in some intermediate value and update `.value` right before notifying subscribers, but it will increase bundle size a bit. Should we?

atom/index.test.ts
psd-coder: Test name was outdated, updated it 👍

atom/index.js
psd-coder: It won't be the same. Empty `batchSeen` means that we are in `batch`, but didn't collect any listener yet. `null` means we aren't in batch.
 
So, it's three states:
- not in batch
- in batch, but no listeners registered yet
- in batch, there are listeners

atom/index.d.ts
psd-coder: Updates won't be batched 🤣 Updated `batch` JSDoc telling this.

atom/index.js
AleksandrSl: Ah, true, sad

atom/index.d.ts
AleksandrSl: I don't know, it's discouraged to use .value anyway, and it could be an escape hatch for people who want to get  updates straight away

atom/index.test.ts
psd-coder: [Added it](https://github.com/nanostores/nanostores/pull/399/changes/d1ba667ea5d567fae6797c7630eb4d14ec0b4147#diff-d1278cb3b79c1f4075d5187320a2b36e344ac872a7cf412830e0149f91eb56eeR668-R685)

ai:
Sorry, too busy this week. I will review it next week.

atom/index.js
ai: What backward compatibility we will have if we will not have `(!changedKey)` here?

psd-coder: Listener has three arguments: `value`, `oldValue`, and `changed`, where `changed` is a string. We could batch `setKey` calls too, but this way, we would need to accumulate the changed keys and pass an array of them as the third parameter. However, that means we would need to change the existing API and pass an array of changed keys, not a string. For single updates, it would contain an array with a single element. This is the proper way to batch `setKey` calls, but it’s a breaking change, which is why I mentioned backward compatibility.

Example:
```
let $user = map({ age: 0, name: 'A' })

$user.listen((v, ov, changed) => {
  // So what must `changed` have here?
})

batch(() => {
   $user.setKey('name', 'B')
   $user.setKey('age', '10')
})
```

psd-coder: If you think we can include this breaking change in a minor release, let’s do it!

AleksandrSl: Just for fun, you can do changedKeys.join(","), and keep the API, it's cursed though

AleksandrSl: By the way, could we just skip the changedKey in case you batch? It's not breaking change. You have to use batch to get the behaviour difference.

Old code like 
```
$user.setKey('name', 'B')
$user.setKey('age', '10')
```
will still work as before

```
batch(() => {
   $user.setKey('name', 'B')
   $user.setKey('age', '10')
})
```
will batch and `changedKey` will be undefined, so it's an implementation detail, which we may fix in the next major, but for minor keep the compatibility

ai: Right now we have cases when `changedKey` is empty, like on `map.set(newObject)`.

I think we can use this pattern here and move to array in next major.

psd-coder: It makes sense! Pushed a commit with the change.

ai:
Right now there is a name conflict with `batched` function.

But we are planning to deprecate it in this minor release https://github.com/nanostores/nanostores/pull/405

PR #400, opened 2026-05-21
Reset computed cache on cleanStores

### LINKED ISSUE #323: `cleanStores` doesn't reset values of `computed`
Hi

I noticed that when using 

```ts
afterEach(() => {
  cleanStores()
})
```

in my tests, all values wrapped in `computed()` are not reset

Example code:

```ts
export const $getThings = createFetcherStore(
  ['project', 'memberships', $projectId],
  {
    fetcher: () =>
      getThings().then((response) => response.data),
  },
);

export const $getProjectMembershipsWithSSHKeys = computed(
  [$getThings],
  (data) => ({
    ...data,
    data: data.data?.memberships.filter(someFilterFn)
  }),
);
```

ATOM00blue:
## Problem
`cleanStores()` resets stores via their internal `[clean]` method, but `computed()` (and `batched()`) stores never defined one. They cache `previousArgs`, `currentEpoch`, and the inner atom's `value` in a closure, so `cleanStores()` only reset the underlying atom's listeners — the computed value stayed stale between tests.

## Fix
Define `$computed[clean]` (dev/test only, mirroring `atom`'s `[clean]`) to clear `previousArgs`, `currentEpoch`, and the cached `value`, then delegate to the atom's existing clean handler. The value is now recomputed after `cleanStores()`. The dev-only code is tree-shaken from production builds (size-limit unaffected). Added a regression test.

Closes #323

### DIFF
--- a/computed/index.js
+++ b/computed/index.js
@@ -1,4 +1,5 @@
 import { atom, nanostoresGlobal } from '../atom/index.js'
+import { clean } from '../clean-stores/index.js'
 import { onMount } from '../lifecycle/index.js'
 import { warn } from '../warn/index.js'
 
@@ -37,6 +38,16 @@ let computedStore = (stores, cb, batched) => {
   $computed.get = () => {
     set()
     return get()
+  }
+
+  if (process.env.NODE_ENV !== 'production') {
+    let cleanComputed = $computed[clean]
+    $computed[clean] = () => {
+      previousArgs = undefined
+      currentEpoch = undefined
+      $computed.value = undefined
+      cleanComputed()
+    }
   }
 
   let timer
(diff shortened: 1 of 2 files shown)


computed/index.js
```diff
@@ -39,6 +40,16 @@ let computedStore = (stores, cb, batched) => {
     return get()
   }
 
+  if (process.env.NODE_ENV !== 'production') {
+    let cleanComputed = $computed[clean]
+    $computed[clean] = () => {
+      previousArgs = undefined
+      currentEpoch = undefined
```
ai, line 47: Is it a good idea to clean it? What reasons do you have?

PR #405, opened 2026-06-14
Add initial debounce function (WIP)

finleyjb:
This creates a generic debouncer that allows for things like:

```typescript
const inputAtom = atom(1);
const debouncedAtom = timeoutDebounce(inputAtom);

debouncedAtom.listen(console.log)

inputAtom.set(2)
inputAtom.set(3)

// After timeout expires, 3 will be logged
```

### DIFF
--- a/debounce/index.d.ts
+++ b/debounce/index.d.ts
@@ -1,0 +1,6 @@
+import type { ReadableAtom, WritableAtom } from '../atom/index.js'
+
+export function timeoutDebounce<T>(
+  inAtom: WritableAtom<T>,
+  timeout = 0
+): ReadableAtom<T>
--- a/debounce/index.js
+++ b/debounce/index.js
@@ -1,0 +1,27 @@
+let debounce = (inAtom, scheduleFn) => {
+  let outAtom = atom(inAtom.get())
+
+  let previousVal
+
+  inAtom.listen(val => {
+    if (!previousVal) {
+      scheduleFn(() => {
+        outAtom.set(previousVal)
+        previousVal = null
+      })
+    } else {
+      previousVal = val
+    }
+  })
+
+  return outAtom
+}
+
+export const timeoutDebounce = (inAtom, timeout) => {
+  timeout = timeout ?? 0
+  let scheduleFn = callback => {
+    setTimeout(val => callback(val), timeout)
+  }
+
+  return debounce(inAtom, scheduleFn)
+}
--- a/debounce/index.test.ts
+++ b/debounce/index.test.ts
@@ -1,0 +1,19 @@
+import FakeTimers from '@sinonjs/fake-timers'
+import { equal } from 'node:assert'
+import { test } from 'node:test'
+
+import { timeoutDebounce } from './index.js'
+import { atom } from '../atom/index.js'
+
+let clock = FakeTimers.install()
+
+test('debounces on call', () => {
+  let inAtom = atom(1)
+  let debouncedAtom = timeoutDebounce(inAtom, 0)
+
+  inAtom.set(2)
+  inAtom.set(3)
+
+  clock.runAll()
+  equal(debouncedAtom.get(), 3)
+})


ai:
Hi. I like this store. 

But can we move it to a separated repo? We can give it official status in `nanostores` org and promoted in the docs.

If you agree with this plan, create a repo, and next I will give you permissions to move it to the org.

finleyjb:
I’d be ok with that. I’d like to get a sense of what requirements there are in terms of say coding standards, dev dependencies, or other things. I’m guessing there’s a preference for using oxc formatting and linting?

Just glancing at the other repos, it looks like they aren’t doing things like publishing through CI to have provenance attestations or using tools like changesets or semantic-releases for versioning. I’m guessing there’s no issue doing those things?

I certainly have ways I’d answer these questions on my own, but I’d like to ensure I’m being a good citizen in the ecosystem.

ai:
> coding standards

Nothing, exactly. But you can copy linters/structure from other stores like https://github.com/nanostores/persistent

> I’m guessing there’s a preference for using oxc formatting and linting?

It is up to you as maintainer. But I can highly recommend oxc since it is very fast.

> Just glancing at the other repos, it looks like they aren’t doing things like publishing through CI to have provenance attestations

CI provenance was less secure than manual 2FA before.

Now they added Staged Releases and it is safer. We just added our system of using it
https://github.com/nanostores/nanostores/blob/main/.github/workflows/release.yml

> changesets or semantic-releases for versioning

I personally don't like it since I prefer to manually write `ChangeLog.md` since ChangeLog should be written from perspective of user (and commit are from perspective of tool developer).

> I’m guessing there’s no issue doing those things?

No, all choices are up to you.

I will give you a permission to publish with official `@nanostores/` org on npm when the repo will be ready.

ai:
Closing since it is too big for core module

ai:
Also, we could think of moving `batched` to your package too.

finleyjb:
I don’t think moving `batched` would be a good idea. Because we’d probably end up having to either expose some implementation details in the core package or rewriting it. Rewriting may be doable, but my gut instinct is that there might be subtleties that would be difficult to get right.

I’d prefer just deprecating batched and recommending this library as the alternative.

ai:
> I’d prefer just deprecating batched and recommending this library as the alternative.

Good idea too.

ai:
@finleyjb right now your module is a blocker to for the next minor release (since we need to deprecate `batched` and put some other solution in the warning) because we are going to introduce `batch` function.

Can I ask you to prepare this repo this week pleeese? =^_^=

finleyjb:
Is there a timeline you have in mind for that?

-------- Original Message --------

ai:
> Is there a timeline you have in mind for that?

I would like to release new minor version next week (monday?).

But we can postpone `batched` deprecation warning on next minor if you would not be able to prepare the repo.

Right now we need mostly the repo URL to put in the warning text.

finleyjb:
I can have something done later next week, but I’m not comfortable committing to Monday

-------- Original Message --------

ai:
Don't worry. We will do one minor release this week and next one in a few weeks later.

finleyjb:
Ok, here's the initial repo setup: https://github.com/finleyjb/grouping

A couple of other things:

1. I see errors.ts files in all these repos, but I don't see how they're run or used. Is there something I'm missing?
2. Is there maybe a better way we can communicate going forward rather than on this PR?
3. Could I maybe request a code review here and there?

ai:
> I see errors.ts files in all these repos, but I don't see how they're run or used. Is there something I'm missing?

It is type tests for https://github.com/ai/check-dts

> Is there maybe a better way we can communicate going forward rather than on this PR?

Telegram, WhatsApp or email? https://sitnik.es/en/

> Could I maybe request a code review here and there?

I recommend pinning CI actions by https://github.com/azat-io/actions-up to reduce risk of supply chain attack.

Rest is nice.

ai:
An idea about creating `@nanostores/utils` or `@nanostores/helpers` to bring all small smart stores. Do you like this idea more than current approach?

finleyjb:
> It is type tests for https://github.com/ai/check-dts

So this sounds like something that [vitest type tests](https://vitest.dev/guide/testing-types) could do?

>  recommend pinning CI actions by https://github.com/azat-io/actions-up to reduce risk of supply chain attack.

I’m asking if you would be willing to do code reviews on changes I make on my repo. If it makes a difference, I’d also be willing to do code reviews for your changes.

> An idea about creating @nanostores/utils or @nanostores/helpers to bring all small smart stores. Do you like this idea more than current approach?

If I understand correctly, you’re suggesting that we move some of the smaller packages to a bigger package?

I can’t think of any benefits that feel like they’d outweigh the cons, and the benefits I can think of could probably be achieved by other methods.

Pros:

* It would allow for code reuse if we were to actually bring the code for these libraries into the repo itself. But I think that would be a lot of work to do as opposed to using submodules.
* It would mean that we’d have to set up less infrastructure for publishing. But we might be able to achieve that through writing a shared GitHub action.
* It would allow having shared coding standards and having tooling like linters, formatters, and build tools. But it sounds like there’s a question of whether that’s actually something we’d want. And it would mean there’s work coercing each library to use those standards.
* Maybe having just one package could make dependency management easier for users? I’m not sure I see that being a major advantage though.

Cons:

* The biggest issue is that it would create a lot of churn for users. Users would have to switch libraries. This might be painful for users if they’re using a lot of these libraries or if they’re depending on various versions. And in general I think they’d get annoyed.
* It would mean we’d have to deprecate the existing libraries and maintain them for some period of time.
* It would create churn for us. I don’t think there’s any way to do this that doesn’t require some amount of changing things.
* I worry that it could create an attitude of “Just throw it in utils, no big deal” that could lead to bloat and leave us with extra code to maintain. So we’d need to have standards about what would go into such a package.
* Packages named `util` or `helpers` are generally considered a code smell by many people, myself included. So that would give me pause about using a library like this. I suppose you could get around it using a name like `misc` though.

One alternative that could get us most of the pros with fewer of the cons would be to set up a synthetic monorepo. That would essentially mean assembling a monorepo using git submodules.

It’s a compromise between a monorepo that throws everything into one repository and polyrepos that are all separate and have nothing to do with one another. It’s also easier to migrate to in our situation. And it would allow separate packages to create less churn for users.

And having code all in one place makes it easier for people using AI.

I know you said you’re not a fan of tools like `changesets`, but I think it would be very helpful for doing the versioning in scenarios like this. And it allows for a separation of the commit message that goes in git history and the message that goes in the changelog.

ai:
> So this sounds like something that [vitest type tests](https://vitest.dev/guide/testing-types) could do?

Yes!

> I’m asking if you would be willing to do code reviews on changes I make on my repo. 

No

> If I understand correctly, you’re suggesting that we move some of the smaller packages to a bigger package?

No, the idea to have 5-8 stores instead of 2-3 (like `debounce`/`throttle`).

For instance, to have one npm package with all simple 1-3 lines smart stores which general purpose.

finleyjb:
> No

Ok. Thanks for being direct.

I’m still not following your question. What I’m gathering is that that this is less about the number of packages and more about the number of stores? Are you suggesting more stores per package? That doesn’t sound like a bad idea.

I can see the benefit of having all 1-3 line stores in a single package. I guess to me it would be a question of how many such stores there are. If there’s a large number of them, then a separate package makes sense. If it’s just a couple, maybe it would just be easier to put them in the main Nanostores package.

-------- Original Message --------

PR #408, opened 2026-06-24
Support nested mutations (deepMap) is listenKeys

Lifo123:
Following the discussion and implementation in [nanostores/react#44](https://github.com/nanostores/react/pull/44).
The listener now evaluates both the exact mutated path and its extracted **base key**:
1. Evaluates `keysSet.has(changed)` for exact matches.
2. Evaluates `keysSet.has(changed.split(/\.|\[/)[0])` to extract the base key of the mutation (e.g., `a.b` extracts to `a`, and `a[0]` extracts to `a`).

I also keep previous PR [#404](https://github.com/nanostores/nanostores/pull/406)

```js
//PR
if (changed === undefined
     - : (keysSet.has(changed)
     + : (keysSet.has(changed) || keysSet.has(changed.split(/\.|\[/)[0]))
) {
}
```

### DIFF
--- a/listen-keys/index.js
+++ b/listen-keys/index.js
@@ -4,7 +4,7 @@ export function listenKeys($store, keys, listener) {
     if (
       changed === undefined
         ? keys.some(key => value[key] !== oldValue[key])
-        : keysSet.has(changed)
+        : (keysSet.has(changed) || keysSet.has(changed.split(/\.|\[/)[0]))
     ) {
       listener(value, oldValue, changed)
     }
--- a/README.md
+++ b/README.md
@@ -253,6 +253,15 @@ listenKeys($profile, ['name'], (value, oldValue, changed) => {
 })
 ```
 
+Nested mutations can be listened to using a base key. For example, listening to `user` will trigger on `user.name` mutations:
+
+```ts
+// Mutating user.name triggers the listener
+listenKeys($state, ['user'], (value, oldValue, changed) => {
+  console.log(`user changed: ${changed}`)
+})
+```
+
 `subscribeKeys(store, keys, cb)` in contrast with `listenKeys(store, keys, cb)`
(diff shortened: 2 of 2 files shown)


listen-keys/index.js
```diff
@@ -4,7 +4,7 @@ export function listenKeys($store, keys, listener) {
     if (
       changed === undefined
         ? keys.some(key => value[key] !== oldValue[key])
-        : keysSet.has(changed)
+        : (keysSet.has(changed) || keysSet.has(changed.split(/\.|\[/)[0]))
```
ai, line 7: Oh, now I see that instead of checking `a.b` we just check for `a` and ignore `b`.

Is it a good thing? People will expect only `a.b` changed, but will get `a.c`, etc?

Maybe we should add to docs here and in React intergration to use base key for this case?

Lifo123: You right, Extracting a base breaks granular reactivity.

To fix this i use a dynamic `startsWith` for both cases `.` `[]` check instead of splitting the string.

```js
keys.has(changed) || 
keys.some(key => changed.startsWith(`${key}.` || changed.startsWith(`${key}[`
```
With this new code first validate is exact path `a.b.c` and if is listen any parent enter in validation key + `.` or `[`.

For documentation i think to add a before I gestión add listenkeys, suggesting using a parent or a exact path for reactivity.

ai: Yes, let’s just add to the docs, I do not think that this change will have a good DX on big scale where user will not undertand the whole system.

Lifo123: So, we maintain the previous logic and add it to the docs?

ai: > So, we maintain the previous logic and add it to the docs?

Yes. Please update the PRs.

PR #410, opened 2026-07-08
fix: correct JSDoc typos in deep-map/path.d.ts

maxtaran2010:
## Summary

Three typos in JSDoc comments inside `deep-map/path.d.ts`:

- `splitted` → `split` in `@param path` description for `getPath` (line 112)
- `splitted` → `split` in `@param path` description for `setPath` (line 134)
- `@retunts` → `@returns` in `setByKey` JSDoc tag (line 159)

"Splitted" is a non-standard form — the correct past participle of "split" is "split". The `@retunts` tag is a plain typo that would cause tools relying on JSDoc tags (typedoc, IDE hover docs) to silently ignore the return type description.

No logic changes — comments/documentation only.

### DIFF
--- a/deep-map/path.d.ts
+++ b/deep-map/path.d.ts
@@ -109,7 +109,7 @@ export type BaseDeepMap = Record<string, unknown>
  * ```
  *
  * @param obj Any object.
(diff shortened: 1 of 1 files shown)


ai:
Thanks!

PR #417, opened 2026-08-12
Keep listeners running when one of them throws

psd-coder:
## Overview

Fixes a bug where one throwing listener permanently stops every listener in the
application. Store values keep updating, so nothing looks broken until you
notice the UI has stopped moving.

## Problem Statement

`drainQueue` exits before `listenerQueue.length = 0` runs when a listener
throws. The queue stays non-empty for good. Every later `notify` sees a
non-empty queue, decides a drain is already in progress, pushes its listeners
and returns. From that point on no listener anywhere fires again.

```js
const $a = atom(0)
const $b = atom(0)

$a.listen(() => { throw new Error('boom') })
$b.listen(value => console.log('b listener saw', value))

try { $a.set(1) } catch {}

$b.set(2)
console.log('$b.value is', $b.value) // 2, correct
// the $b listener never ran, and never will
```

A page in this state cannot be recovered from user code. `batch(() => {})` does
call `drainQueue` from its `finally`, but the loop restarts at index 0, hits the
same entry and throws again. Unsubscribing the bad listener does not help
either: the unbind loop in `listen` starts at `lqIndex + QUEUE_ITEMS_PER_LISTENER`,
so it can never remove the entry at `lqIndex`, which is the one that threw.

The listener that throws belongs to the application, not to Nano Stores. But the
punishment lands somewhere else: an unrelated store goes quiet, values still look
right in the console, and nothing points back at the real throw. Anything that
listens broadly, such as a logger or a devtools bridge, is the most likely
casualty of a throw it had nothing to do with.

## Solution Approach

- `drainQueue` now runs each listener inside its own `try`, so one bad listener
  no longer stops the others. This matches what DOM events and `EventEmitter` do.
- The queue is always cleared, which restores the invariant that `notify` relies
  on. The application recovers on the next change.
- The first thrown value is rethrown once the queue is drained, so the error
  still reaches the caller of `set()`. It is held in a one-element array so that
  a falsy thrown value, for example `throw 0`, is not swallowed.
- `Atom` grows from 340 to 368 B and `Popular Set` from 862 to 891 B.

## Breaking Changes

No API changes. Two behavior changes follow from the fix:

- Listeners queued after a throwing one now run. Before, they were skipped.
- The error reaches the `set()` caller after the queue is drained instead of in
  the middle of it. When several listeners throw, the first thrown value is
  rethrown and the others are dropped.

## Testing

Six tests were added to `atom/index.test.ts`, each covering one part of the
contract:

- an unrelated store keeps working after a listener throws (the report above)
- listeners queued after a throwing one still run
- the first error is rethrown when several listeners throw
- a falsy thrown value still reaches the caller
- `batch` recovers when a listener throws during the flush from its `finally`,
  which is a different path from the existing test for a throwing batch body
- a `computed` keeps being notified when an earlier listener throws, and
  `get()` returns the correct value

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ A tiny state manager for **React**, **React Native**, **Preact**, **Vue**,
 **Svelte**, **Solid**, **Lit**, **Angular**, and vanilla JS.
 It uses **many atomic stores** and direct manipulation.
 
-- **Small.** Between 340 and 864 bytes (minified and brotlied).
+- **Small.** Between 368 and 891 bytes (minified and brotlied).
   Zero dependencies. It uses [Size Limit] to control size.
 - **Fast.** With small atomic and derived stores, you do not need to call
   the selector function for all components on every store change.
--- a/atom/index.js
+++ b/atom/index.js
@@ -10,19 +10,28 @@ const QUEUE_ITEMS_PER_LISTENER = 4
 // from React), causing each bundle to have its own epoch instance.
 export const nanostoresGlobal = (globalThis.nanostoresGlobal ||= { epoch: 0 })
 
+// One listener throwing must not stop the others, and must not leave the queue
+// dirty: a non-empty queue makes every later notify think a drain is running.
 let drainQueue = () => {
+  // Holds the first thrown value, wrapped so that a falsy one is still rethrown
+  let thrown
   for (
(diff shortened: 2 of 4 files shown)


ai:
Thanks. Released in 1.5.0.

Issue #268, opened 2024-02-09
Equality checks and large data sets


hypeJunction:
I am a working with a large dynamic deep nested map of data. I have read in some of the tickets that there is no batch update for maps and deep maps, because the expectation is that the UI framework should take care of throttling rerenders. That however isn't very efficient - I have thousands of updates within dozens of milliseconds and freezes the app because of all the changes that happen to the map. Each change causes a chain reaction of notifying listeners, which then recompute and notify their listeners, and so it goes. For now, I have switched back to atom, replacing the entire state at once, which solves the freezing problem, but I suspect there is still too much unnecessary recomputation going on. 

If we look at the atom setter, we see this equality check:

```js
    set(data) {
      if ($atom.value !== data) {
        $atom.value = data
        $atom.notify()
      }
    },
```

Working with atoms, I can do some other equality checks from the outside. In case of computed values, however, it seems that there is no way to guarantee that values are identical even though values are equal.

I am wondering what your suggestion would be. I see 2 possible solutions:
1. Add support for batch updating maps before firing notify
2. Extend computed with an option to provide a custom function that would check equality before setting the atom value

ai:
> Extend computed with an option to provide a custom function that would check equality before setting the atom value

[`computed`](https://github.com/nanostores/nanostores/blob/main/computed/index.js) is just syntax sugar on top of listener. You can create your own `computed` with any logic.

I think it is better than making API complex with more options.

> Add support for batch updating maps before firing notify

It is also nice idea, but we need research of API and how to implement it with less code.

hypeJunction:
Tried to rework `computed`, but somehow the TypeScript magic got lost :( There is a considerable performance gain though. 
Will be off next week, so will revisit when I am back from holiday.

PR #418, opened 2026-08-13
Add eq and eqKey for custom value comparison

psd-coder:
## Overview

Stores now keep the value comparison in a field instead of a hard-coded `!==`:
`store.eq(oldValue, newValue)` for `set()`, and `map.eqKey(oldValue, newValue, key)`
for `setKey()`. Both default to `Object.is`. Returning `true` means the values
are the same: no write, no notify, and the old value keeps its identity.

```js
import { dequal } from 'dequal/lite'

let $visibleIds = computed($posts, posts => posts.map(post => post.id))
$visibleIds.eq = dequal
```

## Problem Statement

A computed store builds its value from scratch on every run, so a derived array
or object always gets a new identity. Every store and component that depends on
it updates, even when the content did not change. The owner of the store cannot
prevent this from outside, because `computed` performs the write itself.

Issue #268 asked for the same thing as an option of `computed()` and was closed
with "it is better than making API complex with more options". A field is not an
option: it adds no argument to any factory, no overload to any signature, and
nothing to learn for people who never touch it.

Userland can already do this by replacing `set` on a store, and that stays
possible. Core support adds one thing that patching cannot: the check always
runs last, inside `set`, so it cannot land on the wrong side of an `onSet`
wrapper depending on which code ran first.

Using field approach and not option allowed making it compact: just 8 extra bytes for the atom. eq function as an option was giving around 25-30 bytes.

## Solution Approach

- `atom` gets an `eq` field, used by `set`. Default `Object.is`.
- `map` gets an `eqKey` field, used by `setKey`. Default `Object.is`. The key
  name comes as the third argument, because one function serves every key.
- `eq` is declared on `ReadableAtom`, so computed stores expose it. This is the
  main case, and `readonlyType()` already returns the same writable object.
- `map.set()` uses `eq`, `map.setKey()` uses `eqKey`. One function cannot serve
  both: `set` passes two whole store values, `setKey` passes two values of one
  key, and a comparison written for the object shape would silently drop key
  writes.
- Deleting a key with `setKey(key, undefined)` compares nothing, because
  removing a key changes the shape of the object, not a value.
- `deepMap` gets no `eqKey`, since it is deprecated and goes away in 2.0.

## Breaking Changes

The default comparison moves from `!==` to `Object.is`, which changes two corner
cases:

- writing `NaN` over `NaN` no longer calls listeners; today it calls them on
  every write;
- writing `0` over `-0` now calls listeners; today it does not.

IMO, we can neglect them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ export const Admins = () => {
   - [Maps](#maps)
   - [Lazy Stores](#lazy-stores)
   - [Computed Stores](#computed-stores)
+  - [Value Comparison](#value-comparison)
   - [Effects](#effects)
   - [Batching](#batching)
   - [Map Creator](#map-creator)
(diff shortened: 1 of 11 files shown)


ai:
Thanks. Released in 1.5.0.

PR #419, opened 2026-08-15
fix(listen-keys): compare deep paths on whole-store set

### LINKED ISSUE #416: listenKeys stopped firing on whole-store set() for deepMap dot-path keys (regression from #406)
`listenKeys` stopped firing for `deepMap` consumers watching dot-path keys after #406's fix, since the new whole-store-set comparison reads keys as plain bracket properties, which don't exist on a `deepMap`'s nested value shape.

The current implementation:

```js
export function listenKeys($store, keys, listener) {
  let keysSet = new Set(keys)
  return $store.listen((value, oldValue, changed) => {
    if (
      changed === undefined
        : keysSet.has(changed)
    ) {
    }
  })
}
```

`listenKeys`'s own type signature is generic over any store with `setKey(key, value)`, which explicitly includes `deepMap` (dot-path keys like `'profile.name'`), not just `map` (flat keys). But `value[key]` for a dot-path key does a literal bracket lookup, `value['profile.name']` is `undefined` on an object shaped `{ profile: { name: ... } }`, there's no top-level property with that exact name. So for a `deepMap`, `value[key] !== oldValue[key]` is `undefined !== undefined`, always `false`, for every watched key, on every whole-store `.set()` call.

Reproduced against the real published package (`nanostores@1.4.2`), not just by reading the source:

```js
import { deepMap, listenKeys } from 'nanostores'

const $store = deepMap({ profile: { name: 'old' }, other: 1 })
let events = []
listenKeys($store, ['profile.name'], (value, oldValue, changed) => {
  events.push({ changed, name: value.profile.name })
})

$store.set({ profile: { name: 'new' }, other: 1 })
console.log(events) // []  — expected the listener to fire, profile.name did change
```

`setKey('profile.name', 'new')` (the non-whole-store path) still works fine, since `changed` is the literal string `'profile.name'` there and `keysSet.has(changed)` does a correct set-membership check. It's specifically the whole-store `.set()` path from #406 that regressed for deep keys.

`deepMap` itself is deprecated in favor of `@nanostores/deepmap` and slated for removal in 2.0, but it's still shipped and functional today, so this is a live regression for anyone combining `listenKeys` with `deepMap` and a whole-store `.set()` call, worth checking whether `@nanostores/deepmap`'s own key format needs the same `getPath`-aware comparison if it reuses this same `listenKeys`.

ankit1324:
Closes #416

## Problem

`listenKeys` handles a key differently depending on which branch runs:

```js
changed === undefined
  ? keys.some(key => value[key] !== oldValue[key])          // whole-store set()
  : (keysSet.has(changed) || keysSet.has(changed.split(/\.|\[/)[0]))  // setKey()
```

The `setKey` branch resolves paths by root segment (added in #408), but the whole-store `set()` branch treats the key as a plain property. For a path key such as `'profile.name'`, `value['profile.name']` is `undefined` on a value shaped `{ profile: { name } }` — on both sides — so the comparison is `undefined !== undefined`, always false, for every watched key on every `set()`. The listener never fires.

## Change

When a plain lookup finds nothing on both sides, fall back to resolving the key as a path with the existing `getPath`:

```js
value[key] !== oldValue[key] || getPath(value, key) !== getPath(oldValue, key)
```

Two properties this keeps, which is why the direct comparison stays first rather than being replaced:

- **#406's guarantee is preserved.** The comparison is still per-key and value-based, so a `set()` that leaves the watched keys untouched still does not notify — including a sibling change under the same parent, which a root-segment comparison would have reported as a false positive.
- **Plain keys that contain a dot keep working.** A store with a literal `'a.b'` property is still compared directly; resolving it as a path would have looked up `a.b` and broken it. There is a regression test for exactly this.

`getPath` is already a public export, and the two tracked size limits (`atom`, `map` + `computed`) don't include `listenKeys`, so the budgets are unaffected — both still pass.

## Tests

Three cases added to `listen-keys/index.test.ts`:
- a path key fires on a whole-store `set()` that changes it, and does **not** fire for an unrelated sibling or an identical value;
- an array path (`'list[1]'`) behaves the same;
- a plain store whose key literally contains a dot is unaffected.

The first two fail on `main` and pass with the change; the third passes either way and is there to pin the behaviour the fix must not break.

`pnpm test` is green end to end: 56 tests, types, lint, size limits, and the enforced 100% coverage.

One note for review: the new tests need `deepMap`, which is deprecated in core, so I added `listen-keys/index.test.ts` to the existing `typescript/no-deprecated` override in `oxlint.config.ts` — the same entry `deep-map/index.test.ts` already uses. Happy to drop that if you would rather the test lived in `@nanostores/deepmap` instead.

### DIFF
--- a/listen-keys/index.js
+++ b/listen-keys/index.js
@@ -1,9 +1,18 @@
+import { getPath } from '../deep-map/path.js'
+
 export function listenKeys($store, keys, listener) {
   let keysSet = new Set(keys)
   return $store.listen((value, oldValue, changed) => {
     if (
       changed === undefined
-        ? keys.some(key => value[key] !== oldValue[key])
+        ? keys.some(
+            key =>
+              // `map` keys are plain properties, but a `deepMap` key is a path into
+              // the value, where a plain lookup would compare undefined to undefined
+              // and never report a change.
+              value[key] !== oldValue[key] ||
+              getPath(value, key) !== getPath(oldValue, key)
+          )
         : (keysSet.has(changed) || keysSet.has(changed.split(/\.|\[/)[0]))
     ) {
       listener(value, oldValue, changed)
--- a/listen-keys/index.test.ts
+++ b/listen-keys/index.test.ts
@@ -1,7 +1,7 @@
 import { deepStrictEqual, equal } from 'node:assert'
 import { test } from 'node:test'
 
-import { listenKeys, map, subscribeKeys } from '../index.js'
+import { deepMap, listenKeys, map, subscribeKeys } from '../index.js'
 
 test('listen for specific keys', () => {
   let events: string[] = []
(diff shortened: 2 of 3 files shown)


ai:
Thanks!

ai:
Released in 1.5.0.

PR #420, opened 2026-08-16
Align TypeScript declarations with runtime behavior

edloidas:
I went through the type declarations and fixed the ones that are wrong. Some were wrong from the start, others drifted as the JavaScript changed. In every case the runtime does the right thing and the type describes something else.

Only one commit touches runtime code — a small guard in `listenKeys`. 

Bundle size doesn't move; no commit touches `package.json`.

Each commit is a single fix, with the full reasoning in its message: what was wrong, how I confirmed it, and what I deliberately left alone. The commits are the real description. They're logically independent of each other, but several touch the same declaration files in sequence — so dropping commits off the end is clean, while picking one out of the middle will need conflict resolution. `pnpm test` passes at every one of the seventeen commits, not just the final state.

## The commits, grouped by impact

Oldest first inside each group. Several of these are correct but stricter, so TypeScript will start complaining about code that used to compile. Nearly all of that code was already unsafe at runtime, and the examples in the docs were written the guarded way. The first group is safe for any release; the second can surface new compile errors; the last two are the ones I'd hold for a major.

### Safe or additive — I don't expect anyone to change anything

- `701bf01` **Remove redundant TSConfig options** — dropped options TypeScript 7 already turns on.
- `3db19fc` **Accept readonly source tuples in computed, batched, and effect** — you can now pass a source list declared with `as const`.
- `f6b3301` **Export StoreValues from package root** — the type was unreachable, because `nanostores/computed` doesn't resolve for consumers.
- `4b5c27d` **Remove stale notifyId declaration** — the runtime dropped that export in 2023 and the declaration kept insisting it was there.
- `5a81ffc` **Handle a missing old value in listenKeys** — it no longer throws when a notification arrives without an old value. This is the one runtime change, and the one commit that touches files nothing else does.
- `115a546` **Correct TSDoc that contradicts the runtime** — comments only. Two statements in map and deep-map described behaviour the runtime doesn't have; both predate this branch.

### Might surface errors, but only where the code was already unsound

- `efd62e3` **Fix listenKeys callback types** — the changed key is a single key or nothing, never an array.
- `bb6d509` **Fix map listener changed key type** — the changed key can be missing.
- `24bd94e` **Split the source contract for computed, batched, and effect** — `effect()` accepts anything it can read and listen to; `computed()` and `batched()` need a real Nano Store.
- `33d442c` **Require onMount initializer** — it's mandatory again, which is what the runtime has always assumed.
- `ec54479` **Allow a missing old value in onNotify payloads** — lifecycle handlers can be called without one.
- `9f0c427` **Fix mapCreator return types** — created stores carry `id`, and store extensions have to be declared.
- `de8c2ff` **Fix lifecycle payload changed key type** — for a union value, the changed key can come from any branch.
- `de8c2ff` **Fix lifecycle payload changed keys for deep maps** — deep maps report full paths, not top-level keys.
- `e357203` **Fix map keys hijacked by store extensions** — an extension that declares its own `setKey` no longer replaces the map's real keys. Narrow case; if you've never done that, nothing changes.
- `0ab15cc` **Allow missing values in eq and eqKey** — the comparison callbacks can be called without a previous value.

### The two that reach ordinary, correct-looking code

- `a5b70a2` **Fix map types without initial value** — `map<Value>()` with no argument now gives you `Partial<Value>`, because the store really does start empty. This touches every property read on such a store.
- `b38a2d7` **Allow a missing old value in listen callbacks** — the old value can be missing. This touches every listener that reads its second argument, so it's the widest change here.

---

Merge as is, split across 1.5 and 2.0, or drop what you'd rather not ship — glad to reshape it either way. So it's your call.

#### What I left out

I found more while I was in there, but it's either in deprecated parts (async `computed`, deep maps) or not a type problem at all. I'll open those separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ Store’s listeners will receive third argument with changed key.
 
 ```ts
 $profile.listen((profile, oldProfile, changed) => {
-  console.log(`${changed} new value ${profile[changed]}`)
+  if (changed) console.log(`${changed} new value ${profile[changed]}`)
 })
 ```
 
--- a/atom/errors.ts
+++ b/atom/errors.ts
@@ -51,3 +51,8 @@ let $counter = atom(1)
 $counter.eq = (oldValue, newValue) => oldValue === newValue
 // THROWS is not assignable
 $counter.eq = (oldValue: string, newValue: string) => oldValue === newValue
+$counter.eq = (oldValue, newValue) => {
+  // THROWS is possibly 'undefined'
+  oldValue.toFixed(2)
+  return oldValue !== undefined && oldValue === newValue
+}
--- a/atom/errors.ts
+++ b/atom/errors.ts
@@ -7,6 +7,15 @@ $store.listen(value => {
   value.value = 2
 })
 
+$store.listen((_, oldValue) => {
+  // THROWS is possibly 'undefined'
+  oldValue.value
+  if (oldValue !== undefined) {
+    let value: string = oldValue.value
+    console.log(value)
+  }
+})
+
 let $fnStore = atom<() => void>(() => {
   $fnStore.set(() => {})
 })
--- a/computed/index.d.ts
+++ b/computed/index.d.ts
@@ -1,8 +1,8 @@
 import type { ReadableAtom } from '../atom/index.js'
-import type { ListenableStore, Store, StoreValue } from '../map/index.js'
+import type { Gettable, Store, StoreValue } from '../map/index.js'
 import type { Task } from '../task/index.js'
 
-export type StoreValues<Stores extends readonly ListenableStore[]> = {
+export type StoreValues<Stores extends readonly Gettable[]> = {
   -readonly [Index in keyof Stores]: StoreValue<Stores[Index]>
 }
 
--- a/computed/index.d.ts
+++ b/computed/index.d.ts
@@ -2,14 +2,9 @@ import type { ReadableAtom } from '../atom/index.js'
 import type { AnyStore, Store, StoreValue } from '../map/index.js'
 import type { Task } from '../task/index.js'
 
-export type StoreValues<Stores extends AnyStore[]> = {
-  [Index in keyof Stores]: StoreValue<Stores[Index]>
+export type StoreValues<Stores extends readonly AnyStore[]> = {
+  -readonly [Index in keyof Stores]: StoreValue<Stores[Index]>
 }
-
-type A = ReadableAtom<number>
-type B = ReadableAtom<string>
-
-type C = (...values: StoreValues<[A, B]>) => void
 
 interface Computed {
   /**
@@ -22,8 +17,8 @@ interface Computed {
   /**
    * @deprecated Use `@nanostores/async`.
    */
-  <Value, OriginStores extends AnyStore[]>(
-    stores: [...OriginStores],
+  <Value, OriginStores extends readonly AnyStore[]>(
+    stores: readonly [...OriginStores],
     cb: (...values: StoreValues<OriginStores>) => Task<Value>
   ): ReadableAtom<undefined | Value>
   <Value, OriginStore extends Store>(
@@ -45,8 +40,8 @@ interface Computed {
    *
    * Use `@nanostores/async` for async function.
    */
-  <Value, OriginStores extends AnyStore[]>(
-    stores: [...OriginStores],
+  <Value, OriginStores extends readonly AnyStore[]>(
+    stores: readonly [...OriginStores],
     cb: (...values: StoreValues<OriginStores>) => Task<Value> | Value
   ): ReadableAtom<Value>
 }
@@ -72,8 +67,8 @@ interface Batched {
    * })
    * ```
    */
-  <Value, OriginStores extends AnyStore[]>(
-    stores: [...OriginStores],
+  <Value, OriginStores extends readonly AnyStore[]>(
+    stores: readonly [...OriginStores],
     cb: (...values: StoreValues<OriginStores>) => Task<Value> | Value
   ): ReadableAtom<Value>
 }
--- a/computed/index.d.ts
+++ b/computed/index.d.ts
@@ -1,8 +1,8 @@
 import type { ReadableAtom } from '../atom/index.js'
-import type { AnyStore, Store, StoreValue } from '../map/index.js'
+import type { ListenableStore, Store, StoreValue } from '../map/index.js'
 import type { Task } from '../task/index.js'
 
-export type StoreValues<Stores extends readonly AnyStore[]> = {
+export type StoreValues<Stores extends readonly ListenableStore[]> = {
   -readonly [Index in keyof Stores]: StoreValue<Stores[Index]>
 }
 
@@ -17,7 +17,7 @@ interface Computed {
   /**
    * @deprecated Use `@nanostores/async`.
    */
-  <Value, OriginStores extends readonly AnyStore[]>(
+  <Value, OriginStores extends readonly Store[]>(
     stores: readonly [...OriginStores],
     cb: (...values: StoreValues<OriginStores>) => Task<Value>
   ): ReadableAtom<undefined | Value>
@@ -40,7 +40,7 @@ interface Computed {
    *
    * Use `@nanostores/async` for async function.
    */
-  <Value, OriginStores extends readonly AnyStore[]>(
+  <Value, OriginStores extends readonly Store[]>(
     stores: readonly [...OriginStores],
     cb: (...values: StoreValues<OriginStores>) => Task<Value> | Value
   ): ReadableAtom<Value>
@@ -67,7 +67,7 @@ interface Batched {
    * })
    * ```
    */
-  <Value, OriginStores extends readonly AnyStore[]>(
+  <Value, OriginStores extends readonly Store[]>(
     stores: readonly [...OriginStores],
     cb: (...values: StoreValues<OriginStores>) => Task<Value> | Value
   ): ReadableAtom<Value>
--- a/lifecycle/index.d.ts
+++ b/lifecycle/index.d.ts
@@ -19,7 +19,7 @@ type MapSetPayload<Shared, SomeStore extends Store> =
 type AtomNotifyPayload<Shared, SomeStore extends Store> = {
   abort(): void
   changed: undefined
-  oldValue: StoreValue<SomeStore>
+  oldValue: StoreValue<SomeStore> | undefined
   shared: Shared
 }
 
@@ -27,7 +27,7 @@ type MapNotifyPayload<Shared, SomeStore extends Store> =
   | {
       abort(): void
       changed: keyof StoreValue<SomeStore>
-      oldValue: StoreValue<SomeStore>
+      oldValue: StoreValue<SomeStore> | undefined
       shared: Shared
     }
   | AtomNotifyPayload<Shared, SomeStore>
--- a/lifecycle/index.d.ts
+++ b/lifecycle/index.d.ts
@@ -1,3 +1,4 @@
+import type { AllKeys } from '../atom/index.js'
 import type { MapStore, Store, StoreValue } from '../map/index.js'
 
 type AtomSetPayload<Shared, SomeStore extends Store> = {
@@ -10,7 +11,7 @@ type AtomSetPayload<Shared, SomeStore extends Store> = {
 type MapSetPayload<Shared, SomeStore extends Store> =
   | {
       abort(): void
-      changed: keyof StoreValue<SomeStore>
+      changed: AllKeys<StoreValue<SomeStore>>
       newValue: StoreValue<SomeStore>
       shared: Shared
     }
@@ -26,7 +27,7 @@ type AtomNotifyPayload<Shared, SomeStore extends Store> = {
 type MapNotifyPayload<Shared, SomeStore extends Store> =
   | {
       abort(): void
-      changed: keyof StoreValue<SomeStore>
+      changed: AllKeys<StoreValue<SomeStore>>
       oldValue: StoreValue<SomeStore> | undefined
       shared: Shared
     }
--- a/lifecycle/index.d.ts
+++ b/lifecycle/index.d.ts
@@ -150,5 +150,5 @@ export const STORE_UNMOUNT_DELAY: number
  */
 export function onMount<Shared = never>(
   $store: Store,
-  initialize?: (payload: { shared: Shared }) => (() => void) | void
+  initialize: (payload: { shared: Shared }) => (() => void) | void
 ): () => void
--- a/lifecycle/index.d.ts
+++ b/lifecycle/index.d.ts
@@ -1,5 +1,6 @@
-import type { AllKeys } from '../atom/index.js'
-import type { MapStore, Store, StoreValue } from '../map/index.js'
+import type { MapStoreKeys, Store, StoreValue } from '../map/index.js'
+
+type KeyedStore = Store & { setKey: (...args: any[]) => any }
 
 type AtomSetPayload<Shared, SomeStore extends Store> = {
   abort(): void
@@ -11,7 +12,7 @@ type AtomSetPayload<Shared, SomeStore extends Store> = {
 type MapSetPayload<Shared, SomeStore extends Store> =
   | {
       abort(): void
-      changed: AllKeys<StoreValue<SomeStore>>
+      changed: MapStoreKeys<SomeStore>
       newValue: StoreValue<SomeStore>
       shared: Shared
     }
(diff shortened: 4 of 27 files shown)


tsconfig.json
```diff
@@ -1,15 +1,12 @@
...
     "allowJs": true,
-    "esModuleInterop": true,
     "isolatedModules": true,
     "module": "NodeNext",
-    "moduleResolution": "NodeNext",
     "noEmit": true,
     "noImplicitOverride": true,
     "noUncheckedIndexedAccess": true,
     "skipLibCheck": true,
-    "stableTypeOrdering": true,
```
ai, line 12: Why did you remove this?

tsconfig.json
```diff
@@ -1,15 +1,12 @@
 {
   "compilerOptions": {
     "allowJs": true,
-    "esModuleInterop": true,
```
ai, line 4: Why we have all those `tsconfig.json` changes like `esModuleInterop` and `moduleResolution`?

README.md
```diff
@@ -188,7 +188,9 @@ we have extra special helpers `useStore` to re-render the component on
 any store changes.
 
 Listener callbacks will receive the updated value as a first argument
-and the previous value as a second argument.
+and the previous value as a second argument. The previous value can be
```
ai: Please remove docs changes here and in TSDocs. Just `undefined` in TS types is enough.

map/index.d.ts
```diff
@@ -27,6 +27,20 @@ export type AnyStore<Value = any> = {
...
+/**
+ * Everything `effect()` needs from a source: a value to read and a way to
+ * hear that it changed. Every Nano Store satisfies this, and so can an
+ * adapter around a foreign store.
+ *
+ * `computed()` and `batched()` need more than this and ask for `Store`
+ * instead. They skip recomputing until the global epoch moves, and only a
+ * real Nano Store moves it.
+ */
+export type ListenableStore<Value = any> = {
```
ai: Why do we need extra type and can't use `ReadableAtom`?

atom/errors.ts
```diff
@@ -1,4 +1,6 @@
 import { atom, ReadableAtom, readonlyType } from '../index.js'
+// THROWS has no exported member 'notifyId'
+import { notifyId } from './index.js'
```
ai: We do not need to add tests for removing old types, please remove it

computed/index.d.ts
```diff
@@ -1,16 +1,11 @@
 import type { ReadableAtom } from '../atom/index.js'
-import type { AnyStore, Store, StoreValue } from '../map/index.js'
+import type { Store, StoreValue } from '../map/index.js'
 import type { Task } from '../task/index.js'
 
-export type StoreValues<Stores extends AnyStore[]> = {
-  [Index in keyof Stores]: StoreValue<Stores[Index]>
+export type StoreValues<Stores extends readonly { get(): any }[]> = {
```
ai: Maybe we should create new type for `{ get(): any }` like `Gettable`

lifecycle/index.d.ts
```diff
@@ -79,7 +81,7 @@ export function onSet<Shared = never, SomeStore extends Store = Store>(
 export function onNotify<Shared = never, SomeStore extends Store = Store>(
   $store: SomeStore,
   listener: (
-    payload: SomeStore extends MapStore
+    payload: SomeStore extends KeyedStore
```
ai, line 84: Any reasons for that changes?

listen-keys/index.js
```diff
@@ -7,6 +7,7 @@ export function listenKeys($store, keys, listener) {
       changed === undefined
         ? keys.some(
             key =>
+              oldValue === undefined ||
```
ai, line 10: It is not `Align TypeScript declarations with runtime behavior`.

Why do we need that change?

map/index.d.ts
```diff
@@ -37,11 +51,11 @@ export type BaseMapStore<Value = any> = {
   setKey: (key: any, value: any) => any
 } & WritableAtom<Value>
 
-export type MapStoreKeys<SomeStore> = SomeStore extends {
-  setKey: (key: infer K, ...args: any[]) => any
-}
-  ? K
-  : AllKeys<StoreValue<SomeStore>>
+export type MapStoreKeys<SomeStore> = SomeStore extends MapStore
```
ai, line 45: Can you explain reasons for this change too?

tsconfig.json
edloidas: As I stated in that very first commit, `stableTypeOrdering` was added when this project moved to TS6. But TS6 was a transitional version between TS5 and TS7, and this option defaulted to `false`. Now that the project has moved to TS7, it defaults to `true`, so we no longer need to set it explicitly.
As written in the [release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/#updates-since-5.x-and-new-behaviors-from-6.0), `stableTypeOrdering` is `true` by default, and cannot be turned off.

tsconfig.json
edloidas: While this isn't directly just a types fix, and this PR has overgrown itself a bit, I thought it'd still be worth cleaning up some type-related things. Since we have `moduleResolution: NodeNext`, `module` is already `NodeNext`, and `esModuleInterop` defaults to `true` in TypeScript 7 and isn't going to change.

More to that, sinse version 6 it's now impossible to set `esModuleInterop` to `false`, it was [deprecated](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#deprecated---esmoduleinterop-false-and---allowsyntheticdefaultimports-false).

Though I can understand if you want to keep it explicit here. I'm not a fan of that approach myself, and I've seen enough bloated configs dragging along unnecessary properties from times long past.

BTW, there's a reason why, unlike those, `strict` is kept explicit: there are plans to extend it in the future with potentially more strict rules.

But this is a matter of taste, I guess. Say, and I drop the whole commit.

map/index.d.ts
edloidas: In short, `AnyStore` requires the wrong members, `ReadableAtom` requires seven we never call (including internal props like `lc` or `eq`, while `listen` is incorrect for our usecases). `ListenableStore` has two that are actually used and it's the most correct and narrow type here.

`effect()`, `computed()`, and `batched()` used `AnyStore` for source arrays. 
`AnyStore` is:
```ts
export type AnyStore<Value = any> = {
  get(): Value
  readonly value: undefined | Value
}
```
This is not correct. For example, `effect/index.js` uses `get()` and [`listen()`](https://github.com/nanostores/nanostores/blob/main/effect/index.js#L14). 
At the same time, there's a subtler problem too. `ReadableAtom` [`listen()`](https://github.com/nanostores/nanostores/blob/afce8521c0dc644cdd69141defce2955958456d9/atom/index.d.ts#L73-L78) is typed 
```ts
(listener: (value, oldValue) => void) => () => void
```
And it promises the listener gets _both_ values. `effect` calls `store.listen(run)` with a zero-argument run.

computed/index.d.ts
edloidas: Hmm, I missed that part. Guess it would be better than keeping it inline here. I see that this would overlap with `AnyStore` and `ListenableStore`. Would you mind if we declared them as `*Store = Gettable & { ... }`?

lifecycle/index.d.ts
edloidas: So, `onNotify()` is declared like this:
```ts
export function onNotify<Shared = never, SomeStore extends Store = Store>(
  $store: SomeStore,
  listener: (
    payload: SomeStore extends KeyedStore
      ? MapNotifyPayload<Shared, SomeStore>
      : AtomNotifyPayload<Shared, SomeStore>
  ) => void
): () => void
```

If `SomeStore extends KeyedStore` is true, we treat it as a map payload, **otherwise** as an atom payload. We are specifically interested in the map part, as this fix is related to `DeepMapStore`.

Assume we have `atom(0)`, `map({a, b})`, and `deepMap({a, b: {nested: {deep}}})`.
Then we have `onNotify($store, payload => { ... })`. `payload.changed` in the case of an atom will be `undefined`; for a map, it will be `"a" | "b" | undefined`; and for a deep map... `undefined` (with the old type). But it must be `"a" | "b" | "b.nested" | "b.nested.deep" | undefined` instead.

This is how it's produced in the case of a deep map:

1. `SomeStore = DeepMapStore<{ a: number; b: { nested: { deep: number } } }>`
2. Then it's put into the payload:
```ts
payload: DeepMapStore<T> extends KeyedStore
  ? MapNotifyPayload<never, DeepMapStore<T>>
  : AtomNotifyPayload<never, DeepMapStore<T>>
```
3. Now we evaluate:

**Before:**
```ts
DeepMapStore<T> extends MapStore   // FALSE -> missing `eqKey`
```

**After:**
```ts
DeepMapStore<T> extends KeyedStore // TRUE -> `KeyedStore = Store & { setKey: ... }`
```

4. At this point, you may already get it. The deep map was going the atom payload way.

---

As for the shape of `KeyedStore`, I actually had to pick from several variants. And keeping `MapStore` there was the only incorrect one. It can technically be `KeyedStore`, `BaseMapStore`, `{ setKey: (k, v) }`, or `{ setKey: (...any) }`.

`KeyedStore` is declared as:
```ts
type KeyedStore = Store & { setKey: (...args: any[]) => any }
```

- `(...args: any[]) => any` — this part is the actual point. Since the runtime (in `onSet`) asks exactly one question — `if ($store.setKey)`, whether it exists, and that's it. No value or return type. `(...args: any[]) => any` is exactly the simplest answer. That matters because map and deepMap have genuinely incompatible `setKey` signatures.
- `Store &` is needed to make sure only stores can be passed, not just any shape.

The next question you may ask: _Mikita, but we already have `BaseMapStore`, and it looks almost identical:_
```ts
export type BaseMapStore<Value = any> = {
  setKey: (key: any, value: any) => any
} & WritableAtom<Value>
```

Well, yeah, they are functionally identical on every probe. But... it's not used anywhere in this repo. And it enforces the shape of `setKey`, while not expecting the store to be `ReaddableAtom<Value>`.

Once again, your call.

listen-keys/index.js
edloidas: Yes, it's not directly a type fix. 

In master, `notify(oldValue?: ...)` already declares its old value as optional, so a bare `notify()` is valid and passes `undefined`. But when no key is reported, `listenKeys()` falls back to diffing each watched key against that old value, so on master it throws inside the library, before your listener runs:
```js
let $store = map({ a: 1, b: 2 })
listenKeys($store, ['a'], () => {})
$store.notify()
// TypeError: Cannot read properties of undefined (reading 'a') 
```

`listen()` declared the opposite of `notify()`, with `oldValue` always present. My commit `09e4569` (_Allow a missing old value in listen callbacks_) widened the listen callbacks `oldValue` to `| undefined`, and makes the two agree, and `listenKeys()` is the only place in the library that reads through `oldValue` unguarded.

Shipping the accurate type while that call still crashed seemed a bit wrong. But that's more of an argument for _fixing_ it, not for fixing it _here_. If you want, I can drop or create a new PR for this one.

map/index.d.ts
edloidas: Maybe the less valuable and impactful fix amoung all. Not that I saw this done, but the guard is real.

`map()` takes an optional second type parameter that lets you bolt extra members onto the store:
```ts
export function map<Value extends object, StoreExt extends object = object>(
  value: Value
): PreinitializedMapStore<Value> & StoreExt
```
So the store you get back is an intersection: the real map store + your extra stuff. That's normally harmless: you add a reset() method or whatever.

The bug appears when your extension declares a member the store **already** has. Specifically `setKey`:
```ts
type Base = { a: number; b: string };
type Extension = { setKey: (key: 'ext', value: number) => void };
let $extendedStore = map<Base, Extension>({ a: 1, b: '' });
```

Now the store has two `setKey` declarations — the real one, taking 'a' | 'b', and yours taking 'ext'.

The old `MapStoreKeys` figured out a store's keys by reading them off `setKey` first parameter via `infer`:
```ts
export type MapStoreKeys<SomeStore> = SomeStore extends {
  setKey: (key: infer K, ...args: any[]) => any
}
  ? K
  : AllKeys<StoreValue<SomeStore>>
```

When you infer from a type with more than one call signature, TypeScript picks the last one. Not the union of them — the last.
Here is what keys `infer` gets when there are two `setKeys`:
`Base` — "a" | "b"
`Extension` — "ext"
`Base & Extension` — "ext"
`Extension & Base` — "a" | "b"

 You can see that the key type of the store depends on which side of the `&` extension landed on.
 And `map()` puts`Extension` last, and that last wins:
 ```ts
 export function map<Value extends object, StoreExt extends object = object>(
  value: Value
): PreinitializedMapStore<Value> & StoreExt
```
So any extension's key didn't get added to the map's keys. It _replaced_ them.

File: tsconfig.json (hunk 1 above)
Comment (ai) [reply tsconfig.json:12]: Nice idea, thanks (I will copy it to my other open source)

File: tsconfig.json (hunk 2 above)
Comment (ai) [reply tsconfig.json:4]: Nice idea, thanks (I will copy it to my other open source)

File: map/index.d.ts (hunk 4 above)
Comment (ai) [reply map/index.d.ts]: OK, let’s go to minimal types here

File: computed/index.d.ts (hunk 6 above)
Comment (ai) [reply computed/index.d.ts]: > Would you mind if we declared them as *Store = Gettable & { ... }?

Yes, let's set of types like `ListenableStore`

File: atom/errors.ts (hunk 5 above)
Comment (edloidas) [reply atom/errors.ts]: Removed.

File: computed/index.d.ts (hunk 6 above)
Comment (edloidas) [reply computed/index.d.ts]: As discussed, I added `Gettable` and updated the `AnyStore`, `ListenableStore`, and `StoreValue` types to use it in their declarations instead of duplicating the same shape.

File: README.md (hunk 3 above)
Comment (edloidas) [reply README.md]: Removed.

The one part I kept is `README.md:251`:

```diff
$profile.listen((profile, oldProfile, changed) => {
-  console.log(`${changed} new value ${profile[changed]}`)
+  if (changed) console.log(`${changed} new value ${profile[changed]}`)
})
````

Without the guard, the example no longer compiles once `changed` can be missing — `TS2538: Type 'undefined' cannot be used as an index type.` Though we can still drop it, use `changed!`, or rewrite the line so it doesn't index, whichever you prefer.

Comment (edloidas) [reply README.md]: The added TSDoc is reverted in the related commits. One place I could not revert is `map()` doc block in *"Fix map types without initial value"*. Master has a single signature, and this splits it into two overloads, so one block had to become two. Also, master `@param init` named a parameter that does not exist.

Comment (edloidas) [reply README.md]: Also, pushed one more commit (see the last one), that touches TSDoc only. Specifically `@param changedKey`, in map and deep-map, on both `listen()` and `subscribe()`, e.g.:

```diff
- * @param changedKey Key that was changed. Will present only if `setKey`
- *                   has been used to change a store. It is `undefined` when
- *                   changes were coalesced inside `batch`.
+ * @param changedKey Key that changed, when the notification carries one.
````

**Why I did this and why I think it is correct:**

"Present only if `setKey` has been used" means key present -> `setKey` was called, and that is false:
```ts
$profile.notify(undefined, 'name') // listener receives changed = "name"
```

You see, no `setKey` involved. `notify()` is public API with an optional second argument, and it delivers the key the same way `setKey` does — because `setKey` is implemented as `notify(oldMap, key)` (`map/index.js:13`). Read the other way around, it also fails: inside `batch()`, the key is dropped (`atom/index.js:95`), so `setKey` delivers nothing there.

The list of `undefined` cases was also incomplete — it includes `set()`, a bare `notify()`, anything inside `batch()`, and `subscribe()` first call, which is a plain one-argument `listener(value)` (`atom/index.js:115`). The old text mentioned only `batch`.

The replacement makes one claim that holds in every one of those cases, for maps and deep maps, and for `listen` and `subscribe`. I kept it short on purpose: the accurate version of the old sentence would require four clauses, each describing a different method rather than this parameter. `set()`'s own docs already say `changedKey` will be `undefined`, and the signature says `AllKeys<Value> | undefined`.

And for `eqKey`:

```diff
- * Compares the previous and next values for a key on every `setKey`.
+ * Compares the previous and next values for a key on `setKey`.
...
  * @remarks
- * Deleting a key with `setKey(key, undefined)` does not call it. Whole-value
- * writes use {@link ReadableAtom#eq} instead.
+ * Deleting an existing key skips the comparison and notifies either way.
+ * Whole-value writes use {@link ReadableAtom#eq} instead.
```

Both parts were wrong. `setKey(key, undefined)` skips `eqKey` only when the map already has the key — if the key does not exist, the delete branch is skipped and `eqKey` runs with both values as `undefined`. The delete path also notifies regardless of what a custom `eqKey` would return:

```ts
$mapStore.eqKey = () => true
$mapStore.setKey('x', undefined) // still notifies, without calling eqKey
```

ai:
Thanks. Released in 1.5.1.

PR #423, opened 2026-08-31
fix(computed): detect signed-zero dependency changes

lprnmns:
## Problem

A computed store can remain stale when an atom changes from `-0` to `0`. The atom correctly treats those values as different, but the computed dependency cache compares them with `!==`, so a derived value such as `1 / value` remains `-Infinity` instead of becoming `Infinity`.

## Fix

Compare cached dependency values through each source store's `eq(oldValue, newValue)` function. Its default `Object.is` behavior detects signed-zero changes, while custom equality functions remain effective for computed dependencies. The `every()` form keeps the enforced Popular Set size at its 912 B budget.

## Tests

- `pnpm bnt -t 'signed zero' computed/index.test.ts` — passed
- `pnpm bnt -t 'dependency equality function' computed/index.test.ts` — failed before the follow-up and passed after it
- `pnpm bnt computed/index.test.ts atom/index.test.ts` — passed
- `pnpm test` — passed; 60 tests, full coverage/lint/types/sync checks, Popular Set 912 B / 912 B
- `pnpm exec oxlint computed/index.js computed/index.test.ts` — passed
- `git diff --check` — passed

## Compatibility

This aligns computed dependency caching with the source store's existing equality contract. The default behavior changes only the signed-zero edge case; stores that replace `eq` retain their chosen comparison semantics. No package manifest, lockfile, or public signature changed.

## Related issue

Independent reproduction; no issue linked.

### DIFF
--- a/computed/index.js
+++ b/computed/index.js
@@ -12,7 +12,7 @@ let computedStore = (stores, cb, batched) => {
     if (currentEpoch === nanostoresGlobal.epoch) return
     currentEpoch = nanostoresGlobal.epoch
     let args = stores.map($store => $store.get())
-    if (!previousArgs || args.some((arg, i) => !Object.is(arg, previousArgs[i]))) {
+    if (!previousArgs?.every((arg, i) => stores[i].eq(arg, args[i]))) {
       previousArgs = args
       let value = cb(...args)
       if (value && value.then && value.t) {
--- a/computed/index.js
+++ b/computed/index.js
@@ -12,7 +12,7 @@ let computedStore = (stores, cb, batched) => {
     if (currentEpoch === nanostoresGlobal.epoch) return
     currentEpoch = nanostoresGlobal.epoch
     let args = stores.map($store => $store.get())
-    if (!previousArgs || args.some((arg, i) => arg !== previousArgs[i])) {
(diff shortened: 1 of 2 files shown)


computed/index.js
```diff
@@ -12,7 +12,7 @@ let computedStore = (stores, cb, batched) => {
     if (currentEpoch === nanostoresGlobal.epoch) return
     currentEpoch = nanostoresGlobal.epoch
     let args = stores.map($store => $store.get())
-    if (!previousArgs || args.some((arg, i) => arg !== previousArgs[i])) {
+    if (!previousArgs || args.some((arg, i) => !Object.is(arg, previousArgs[i]))) {
```
ai: Should we use new `store#eq` to be able to replace equality function? https://github.com/nanostores/nanostores/blob/main/atom/index.d.ts#L35-L38

lprnmns: Yes — updated in ec0a59a to compare cached dependency values through each source store eq(oldValue, newValue). The default Object.is behavior still covers signed zero, and a regression now covers custom equality. Full pnpm test passes; Popular Set remains within budget at 912 B / 912 B.

PR #430, opened 2026-09-04
Use Object.is in listenKeys whole-store comparisons

lprnmns:
`listenKeys` filters whole-store notifications with strict inequality even though Nano Stores documents `Object.is` as its default comparison. This makes an unchanged watched `NaN` fire spuriously and misses a watched `-0` to `+0` change.

This uses `Object.is` for both direct-key and deep-path comparisons. Other values keep the same behavior.

Tests:

- `pnpm bnt -t 'Object.is' listen-keys/index.test.ts`
- `pnpm test`
- `git diff --check`

All passed. The full suite reports 66 tests, 100% line coverage, and the Popular Set remains within its 912 B size limit.

### DIFF
--- a/listen-keys/index.js
+++ b/listen-keys/index.js
@@ -11,8 +11,8 @@ export function listenKeys($store, keys, listener) {
               // `map` keys are plain properties, but a `deepMap` key is a path into
               // the value, where a plain lookup would compare undefined to undefined
               // and never report a change.
-              value[key] !== oldValue[key] ||
-              getPath(value, key) !== getPath(oldValue, key)
+              !Object.is(value[key], oldValue[key]) ||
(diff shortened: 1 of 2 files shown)


listen-keys/index.js
```diff
@@ -11,8 +11,8 @@ export function listenKeys($store, keys, listener) {
               // `map` keys are plain properties, but a `deepMap` key is a path into
               // the value, where a plain lookup would compare undefined to undefined
               // and never report a change.
-              value[key] !== oldValue[key] ||
-              getPath(value, key) !== getPath(oldValue, key)
+              !Object.is(value[key], oldValue[key]) ||
```
ai: Should we also use `$store.eq` here? (think how to be sure that `eq` is present without extra check, maybe `peerDependencies`)

